PR #7, opened 2020-02-04
Update pvc.yaml

voslartomas:


sentry/templates/pvc.yaml
```diff
@@ -2,7 +2,7 @@
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
-  name: {{ template "sentry.prefix" . }}sentry-data
+  name: {{ template "sentry.prefix" . }}sentry
```
Mokto: Thanks ! I think we should call this `{{ template "sentry.prefix" . }}pvc`. And then we would need to change the deployment pvc value.

I would expect people to use `sentry-` as a prefix most of the time.

Can also update the version to 0.5.1, so that everyone can use your fix.

Thanks again!

Mokto (changes requested):

sentry/templates/pvc.yaml
voslartomas: Where do you set this sentry.prefix??

Mokto: https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/_helper.tpl#L6 here

voslartomas: Thanks, I've updated PR.

sentry/values.yaml
```diff
@@ -1,4 +1,4 @@
-prefix:
+prefix: sentry-
```
Mokto: Sorry if it's not clear, but the `-` is already added automatically

Mokto (changes requested):

sentry/values.yaml
voslartomas: Got it :)

PR #19, opened 2020-02-10
Add ingress.

### LINKED ISSUE #8: Add an ingress resource

voslartomas:
Closes #8

### DIFF
--- a/sentry/templates/ingress.yaml
+++ b/sentry/templates/ingress.yaml
@@ -19,7 +19,7 @@ spec:
         paths:
           - path: {{ default "/" .Values.ingress.path | quote }}
             backend:
-              serviceName: {{ template "sentry.fullname" . }}
+              serviceName: {{ template "sentry.prefix" . }}sentry-web
               servicePort: {{ .Values.service.externalPort }}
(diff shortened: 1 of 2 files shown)


sentry/templates/ingress.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.ingress.enabled -}}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+ name: {{ template "sentry.prefix" . }}-sentry
```
Mokto: Shouldn't it be `{{ template "sentry.prefix" . }}sentry` ?

sentry/values.yaml
```diff
@@ -172,6 +172,9 @@ slack: {}
 #   clientSecret:
 #   verificationToken:
 
+ingress:
+  enabled: false
```
Mokto, line 176: Could you add all the possible values please ? (You could comment them for now as the ingress is disabled per default)

Mokto (changes requested):

sentry/values.yaml
voslartomas: Done

PR #24, opened 2020-03-04
fix: allow Redis only and fix Redis password use in BROKER_URL

Kampfmoehre:
This allows to skip the deployment of RabbitMQ and therefore skip the deployment of the Sentry worker that connects to it.

This also fixes the Sentry Redis Worker, so that the correct BROKER_URL is used, when Redis is configured to connect with password.

This also updates the Sentry configmap to set the Broker URL to a Redis connection string, if RabbitMQ is disabled.


Mokto:
Amazing! Thanks.

I'll put my comments right away

sentry/templates/configmap-sentry.yaml
```diff
@@ -161,7 +161,16 @@ data:
     # information on configuring your queue broker and workers. Sentry relies
     # on a Python framework called Celery to manage queues.
     
+    {{- if .Values.rabbitmq.enabled }}
     BROKER_URL = os.environ.get("BROKER_URL", "amqp://{{ .Values.rabbitmq.username }}:{{ .Values.rabbitmq.password }}@{{ .Values.rabbitmq.host }}:5672//")
+    {{- else if .Values.redis.enabled }}
+      {{- if .Values.redis.password }}
```
Mokto: This is repeated, could you use a template helper instead?

Thanks.

sentry/templates/deployment-sentry-cron.yaml
```diff
@@ -61,7 +61,15 @@ spec:
           - name: C_FORCE_ROOT
             value: "true"
           - name: BROKER_URL
+            {{- if .Values.rabbitmq.enabled }}
             value: amqp://{{ .Values.rabbitmq.username }}:{{ .Values.rabbitmq.password }}@{{ .Values.rabbitmq.host }}:5672
+            {{- else if .Values.redis.enabled }}
```
Mokto: `{{- else }}` is enough here. One of those is required anyways

sentry/templates/deployment-sentry-worker-redis.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.redis.enabled }}
```
Mokto: `if not .Values.rabbitmq.enabled`

sentry/values.yaml
```diff
@@ -136,11 +136,15 @@ postgres:
...
 rabbitmq:
+  # either use RabbitMQ or Redis as broker
+  enabled: true
   username: guest
   password: guest
   host: rabbitmq-ha
 
 redis:
+  # either use RabbitMQ or Redis as broker
+  enabled: false
```
Mokto: This implies that Redis might not be needed.
Redis is always needed.

You can remove that and put `if not .Values.rabbitmq.enabled` as a condition.

Mokto (changes requested):

sentry/templates/deployment-sentry-cron.yaml
Kampfmoehre: fixed

sentry/templates/deployment-sentry-worker-redis.yaml
Kampfmoehre: fixed

sentry/values.yaml
Kampfmoehre: Removed

sentry/templates/configmap-sentry.yaml
Kampfmoehre: I changed this, since Redis is always enabled. Since I'm fairly new to helm charts, I'm not sure how to write helpers yet. But I could take a look at it.

Mokto:
Thanks ! That's perfect. Could you just activate rabbitmq per default, please ?

Kampfmoehre:
Of course.
Btw: I thought about putting password in secrets, did you start any work on this yet?

Mokto:
No, not yet ! Go for it ;) 

I'll merge your PR in the early evening.

Thanks for your work !

PR #27, opened 2020-03-06
use secrets

Kampfmoehre:
Beginning of the work to put all password stuff in Secrets.

I started with the BROKER_URL which can contain RabbitMQ or Redis password. I put that into a secret and mount it into the Sentry pods that need it.

I also tried to use the Redis password from a secret instead of providing it in the values. This works for the Snuba pods where I can set them to ENV variables.

[The old chart](https://github.com/helm/charts/blob/master/stable/sentry/templates/configmap.yaml) states, that Redis password that is normally set in sentry.conf.py can be set via env variable, but I have no idea how the update the BROKER_URL when no password is given via the values.

I am fairly new to helm charts, so my helm template fu is not enough. @Mokto is there a way to read a value from an existing secret and use that in a template?


donbeave:
Good job!
@Mokto any change to merge this PR?

Kampfmoehre:
Right now it is a little bit difficult and maybe not the best time to merge, because if you specify no Redis password in the values but do not enable RabbitMQ, the BROKER_URL env would be wrong.

I would like to store the Redis Passwort in a user created secret instead of providing it with the chart values, but for the BROKER_URL in the sentry.conf.py file we need to know the password. And as far as I know, there is no possibility to read it from an existing secret during helm install.

What I can imagine right know:
- Create an init container that reads the password from secret, builds the BROKER_URL and writes that to sentry.conf.py and put that in another secret instead of a config map
- Let the user write the BROKER_URL in a secret themselves. But I do not know, if we can omit this in the config file, after all it is set in some ENV variables directly for some pods.

Both of those sound to appealing for me. Becuase of the amount of passwords that are stored in the sentry.conf.py file, it would generally be good, to put that in a secret rather than a config map. But maybe we can remove some of those sensible values from the file and mount them from secrets into env variables.

Mokto:

Kampfmoehre:
Ah sorry, completely forgot about this one. We moved Sentry out of the cluster a long time ago since it made too many problems with performance. if anyone still needs this, feel free to copy any progress made here.

PR #33, opened 2020-03-30
Fixing 9 -> 10 Upgrade Path (From legacy chart). Numerous small fixes to Chart.

DandyDeveloper:
Fixes include: 
 - Better Broker and Redis support with dependencies
 - Using official Kafka repository and settings
 - Increase worker count to 3 for typical Sentry workload
 - More inline changes with postgresql dependency expectations.

**Problems**
Right now, biggest issue is without a Helm plugin, there's no nice way to control the installation of all these components, so you end up with a couple of pod restarts (max 2 in my experience) before everything comes up successfully. 

I've confirmed and on an initial install. This works without problems, I am testing the upgrade path and may have some additional minor updates required.


DandyDeveloper:
@Mokto I've made a bunch of updates here. It's a lot to take in. I've noted the key things in the description.

Mokto:
Can you please leave the chart dependencies as zip files, as it should ?

It is also going going easier for me to read you PR.

Thanks

DandyDeveloper:
@Mokto I have added them as requested - Quick question: Why do you want them in there? Seems like unnecessary data in the repo?

Mokto:
Ah, thanks! You're right, you can remove them if you want. 
I was just struggling with so many charts files ;)

I was also thinking, maybe it's time to include :
- redis as highly available (through redis-ha + haproxy)
- kafka as highly available. (See https://github.com/sentry-kubernetes/charts/pull/26)

I'm aware that the current stable chart is using `redis`, not `redis-ha`, but for a migration, it's still ok, I guess.

I would put the whole sentry stack as highly available.

What do you think ?

DandyDeveloper:
@mokto I own the redis-ha chart so, why not :). https://github.com/dandydeveloper/charts

As for Kafka, maybe in a separate PR? Not that familiar with the Kafka architecture for ha. But I definitely think it's a good idea.

Mokto:
Ok, let's do that then. 

What you've done is really nice, but it's a big change for anyone using the repo.
So let's do all the big changes at once.

And I'll change Kafka after your pull request is merged.

Mokto:
By the way, have you tested the upgrade path?

DandyDeveloper:
@Mokto Not fully, testing edge cases that I can think of.

DandyDeveloper:
@Mokto Having some issues with the Redis upgrades, looks like it is trying to update parts of the statefulset. Needs a little more time.

DandyDeveloper:
@Mokto All done, I've tested a few use cases, added a README for the edge cases basically. Overall, it works a treat. 

I'd worry about HA redis and PSQL outside of this, because right now, with the helm 2 to 3 being so jumpy, it makes upgrading several components really frustrating.

For some reason, linting is failing... Not sure why?

Mokto:
Still in the process of checking.

I've made a few changes to support external databases.
And fix the lint issue.

Mokto:
I'm almost done ;) I'll merge your PR soon (thanks!). And I'll create a PR immediately after to fix the small issues.

sentry/values.yaml
```diff
@@ -107,54 +98,33 @@ hooks:
...
 
 system:
-  url: ''
-  adminEmail: ''
+  url: ""
+  adminEmail: ""
   secretKey: 'icLq77rCyY_qrMMpXa6TQNjkDV6mU!c'
   public: false #  This should only be used if you’re installing Sentry behind your company’s firewall.
 
 mail:
```
gabemeola, line 106: Should this key stay `email` for interop with the old chart?

sentry/values.yaml
```diff
@@ -245,6 +215,81 @@ filestore:
 
 config:
   yaml: |
-    # YAML
```
gabemeola, line 248: Similar question. Old chart calls this `configYml` and `sentryConfPy`

sentry/values.yaml
```diff
@@ -107,54 +98,33 @@ hooks:
...
-  db: 0
-
 kafka:
   host: kafka-confluent
   port: 9092
 
 service:
   name: sentry
-  type: LoadBalancer
+  type: ClusterIP
```
gabemeola, line 127: Curious, why change from `LoadBalancer`?

i5okie: Not everybody uses LoadBalancer. It's an expensive option. I think it's more common to see some sort of an ingress controller being used with service set to ClusterIP. IMO anyway.

gabemeola: Makes sense thanks for clarifying

sentry/values.yaml
Mokto: Thanks for your report. I'll take care of that in the next PR

sentry/values.yaml
Mokto: Thanks for your report. I'll take care of that in the next PR

sentry/values.yaml
Mokto: Actually this one is not the previous chart.

sentry/values.yaml
DandyDeveloper: The good thing is upgrade paths deal with this. It's less of an issue as config maps are different anyway. So it's not a huge problem for people to change the values.

PR #35, opened 2020-04-03
feat: kafka ha

Mokto:


Mokto:
@DandyDeveloper I had to revert your change that removed external charts.

It was not including them when working with the sentry chart as a local dependency.

```
dependencies:
  - name: sentry
    version: 0.10.0
    repository: file://../charts/sentry
```

DandyDeveloper:
@Mokto I'm not sure if that necessarily means adding them back: 

The CI will automatically fetch and publish with the charts. Local dependency won't work, but that would be expected.

That being said, entirely your choice. I don't think it impacts much either way.

Mokto:
Yes, I guess I'll keep them ;) 

So it works for both local dev & packaging.

Thanks!

PR #37, opened 2020-04-06
Fixing indentation on spec.env for Deployments

DandyDeveloper:
Indentation was incorrect which meant adding envs would break the chart templating.

I've sneaked in a fix on the configmap that I actually broke (sorry)


DandyDeveloper:
@Mokto Need some thumbs,

PR #42, opened 2020-04-06
feat: hpa for snuba

Mokto:
@DandyDeveloper I'll just merge this one and the 0.11 will be deployed

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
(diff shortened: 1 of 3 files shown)


DandyDeveloper:
@Mokto Does Snuda need this? I don't think it's used that excessively. Isn't 3 already pretty overkill?

Mokto:
I'm not sure, but our needs might be pretty low compared to other people.
That's optional, disabled per default, I don't see any issue with that !

We could maybe set 2 containers per default, if you think 3 is overkill.

Mokto:
Merged it as the rest is needed to fix the current issues we have

PR #43, opened 2020-04-06
fix: kafka env if disabled

Mokto:
I had to kill and restart my Kafka instance (after a high load during the weekend, the snuba consumer didn't want to start anymore).

I realized it was not possible to set kafka to disabled.

### DIFF
--- a/sentry/templates/_helper.tpl
+++ b/sentry/templates/_helper.tpl
@@ -191,7 +191,7 @@ Set Kafka Confluent host
 {{- if .Values.kafka.enabled -}}
     {{- template "sentry.kafka.fullname" . -}}
 {{- else -}}
-    "kafka-confluent"
+    kafka-confluent
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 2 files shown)


sentry/templates/_helper.tpl
```diff
@@ -191,18 +191,18 @@ Set Kafka Confluent host
...
 {{- end -}}
 {{- end -}}
 
 {{/*
 Set Kafka Confluent port
 */}}
 {{- define "sentry.kafka.port" -}}
-{{- if .Values.kafka.enabled -}}
-    {{- default "9092" .Values.kafka.service.port }}
+{{- if and (.Values.kafka.enabled) (.Values.kafka.service.port) -}}
```
DandyDeveloper, line 202: Don't we have this as a default anyway in the chart? So even if it's missing it'll use the parent value and therefore will always be present?

DandyDeveloper (approved):
Minor comment, just incase as I'm not 100% sure.

sentry/templates/_helper.tpl
Mokto: Theoretically yes, but for some reason it wasn't working without it.

I'll try to apply it again before merging the PR

PR #46, opened 2020-04-08
SENTRY_EVENT_RETENTION_DAYS should be interpreted as int.

DandyDeveloper:


sentry/templates/configmap-sentry.yaml
```diff
@@ -142,7 +142,7 @@ data:
     # and thus various UI optimizations should be enabled.
     SENTRY_SINGLE_ORGANIZATION = True
 
-    SENTRY_OPTIONS["system.event-retention-days"] = env('SENTRY_EVENT_RETENTION_DAYS') or 90
+    SENTRY_OPTIONS["system.event-retention-days"] = int(env('SENTRY_EVENT_RETENTION_DAYS') or 90)
```
Mokto, line 145: Not a python expect but shouldn't `or 90` be outside the parenthesis?

DandyDeveloper: Nope, try it locally: 

```python
>>> number = int("") or 90
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: ''
>>> number = int("" or 90)
>>> print(number)
90
```

Mokto: Definitely I'm not a python expert ahah

Mokto (approved):

PR #48, opened 2020-04-09
fix: worker should work with both redis & rabbitmq

Mokto:


DandyDeveloper (approved):
Sorry this one slipped through!

PR #50, opened 2020-04-11
feat: add pvc to clickhouse

Mokto:
Context : after a Kubernetes upgrade I lost all my tables


Mokto:
Sadly the clickhouse chart doesn't allow us not to use to storageClass.

sentry/values.yaml
```diff
@@ -234,6 +234,14 @@ clickhouse:
   enabled: true
   clickhouse:
     imageVersion: "19.16"
+    persistentVolumeClaim:
+      enabled: true
+      dataPersistentVolume: 
+        enabled: true
+        accessModes: 
+        - "ReadWriteOnce"
+        storageClassName: "standard"
```
DandyDeveloper: Common practice is to set this as `"default"`. May be better?

DandyDeveloper (approved):
Suggestion, but not essential.

sentry/values.yaml
Mokto: It's unclear in the k8s docs, but I think it's only when you don't specify "storageClassName" that default is used. Could you confirm?

DandyDeveloper: Sorry completely missed this reply... Oops. Basically no "baremetal" cluster comes with a storage class, so blank typically defaults to `default` in my experience... Beyond that though really it's down to the user. I don't think it's a big deal either way.

Mokto:
@DandyDeveloper I removed the storageClass per default, by cloning the clickhouse chart inside our repo.

DandyDeveloper:
> @DandyDeveloper I removed the storageClass per default, by cloning the clickhouse chart inside our repo.

:+1:

PR #55, opened 2020-04-15
Fix: pass postgresql password to user-create job

telecatser:
Additional fix for https://github.com/sentry-kubernetes/charts/pull/52

PR #58, opened 2020-04-17
fix: do not complete the user-create on error

wilmardo:
This prevents the user-create job to just complete and get removed as complete Job even when the user wasn't created due an error.
I ran into #55 and couldn't login but the Helm chart install completed nicely. The user-create job must have thrown an error but it exited 0 due the `|| true`. This I couldn't check since the Job had completed successfully and was already removed by Helm.

This uses the new exit code `3` in the user create script introduced in getsentry/sentry#16440 to just exit 0 when the user already exists.
All the other (general) errors like, I couldn't connect to the DB are still thrown and will show up as an errored Job in Kubernetes.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 0.12.2
+version: 0.12.3
 appVersion: 10.0.0
 dependencies:
   - name: redis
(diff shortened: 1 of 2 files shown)


Mokto:
Really nice.

I didn't know about this fix, and it was pretty annoying for me, and probably a few others ;) 
Would you care bumping the chart patch version ?

Thanks!

Mokto (approved):

wilmardo:
@Mokto Thanks, I was annoyed too so I proposed a fix upstream haha

Chart is one patch version bumped to 0.12.2

Mokto:
Thanks but we're already in 0.12.2 ;)

wilmardo:
@Mokto Woops fork was lagging behind :neckbeard: 
Now updated to 0.12.3

PR #70, opened 2020-05-21
Run Snuba in Distributed Dataset Mode

### LINKED ISSUE #68: Events are missing
Hi,
wondering if anyone encountered similar issue and might know what might be causing this.
Sentry receives event, sends email about event being created. But, if you follow the link to sentry, it says:
`Sorry, the events for this issue could not be found.`
Even though it shows: 
`Events: 1`
but if you click on `Events` tab, it is also empty.

It happens to some events only. At some point something changes, and then events are coming through and showing up just fine.

Nothing in logs shows any errors or anything that would explain this. 
snuba consumer is working fine and no errors.
workers are working fine and not backlogged. 
Queue is completely empty.

Only thing that is different from default chart, is that we don't use rabbitmq, instead using redis as broker. Could that cause such behavior? 

Any help or ideas would be greatly appreciated.

J0sh0nat0r:
This pull request configures Snuba to run in distributed dataset mode by:

- Running `snuba-migrate` and `snuba-db-init` jobs against each Clickhouse node
- Configuring Snuba via a `settings.py` file, defaulting `DATASET_MODE` to `distributed`
- Adding a job to create the `sentry_dist` table on each Clickhouse node

Something of note: I've removed the multiple attempts for both the `snuba-migrate` and `snuba-db-init` jobs as it seems the referenced issue has been closed, and it seems to work locally however I though I should mention it just in case.

Fixes #68

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 1.0.0
+version: 1.1.0
 appVersion: 10.0.0
 dependencies:
   - name: redis
--- a/sentry/templates/configmap-sentry.yaml
+++ b/sentry/templates/configmap-sentry.yaml
@@ -1,5 +1,5 @@
-{{- $redisPort := .Values.redis.redisPort -}}
-{{- $redisHost := .Values.redis.hostOverride | default (include "sentry.redis.host" .) -}}
+{{- $redisHost := include "sentry.redis.host" . -}}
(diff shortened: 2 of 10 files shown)


Mokto:
Amazing, thanks !

```
Something of note: I've removed the multiple attempts for both the snuba-migrate and snuba-db-init jobs as it seems the referenced issue has been closed, and it seems to work locally however I though I should mention it just in case.
```
You're right, the issue has been fixed since then.

I'll check your code this morning.

Mokto:
Good job, I'll merge it now ;)

PR #73, opened 2020-05-22
feat: create a unique id to allow restarting jobs manually

Mokto:


DandyDeveloper (changes requested):
Why don't we employ a technique to generate a sha256 based on whether there are changes to an object in certain files throughout. See: https://helm.sh/docs/howto/charts_tips_and_tricks/

Specifically: **Automatically Roll Deployments**

We just have an annotation generate a sha256 for each template file using ` .Files.Get`. It's flawless and always works that way.

Mokto:
That's what we use already for configmaps, secrets, etc...
In that case it's a bit different. Let's say you've lost some tables for some reason.

You haven't made any config change but you still want to re-run your jobs.

DandyDeveloper:
@Mokto Just to be clear, in my experience, every time I run upgrade anyway the jobs here are ran? 

That being said, there has to be a better way to do this. 

How does this differ from just having `.Values.hooks.enabled` set to false?

Mokto:
I'm using Pulumi which is not working like this. But maybe i should take of this specific use case myself.
With Pulumi the job is detected as the same as before, so it's not runned everytime.

My only way, right now, is to do 2 deployments: with enabled = false and then with enabled = true.

Mokto:
I'll close it.

PR #74, opened 2020-05-22
fix hpa

pservit:
- Add hpa for workers
- Do not set replicas for deployment when autoscaling enabled, repicas count managed by hpa

### DIFF
--- a/sentry/templates/deployment-sentry-web.yaml
+++ b/sentry/templates/deployment-sentry-web.yaml
@@ -13,7 +13,9 @@ spec:
         app: {{ template "sentry.fullname" . }}
         release: "{{ .Release.Name }}"
         role: web
+{{- if not .Values.sentry.web.autoscaling.enabled }}
   replicas: {{ .Values.sentry.web.replicas }}
+{{- end }}
   template:
     metadata:
       annotations:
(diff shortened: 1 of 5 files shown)


Mokto:
Thanks for your work.

1. About autoscaling for workers, I didn't include it myself because I think the cpu metric is not the right one to use. In my case I've used prometheus adapter and scale based on the size of the rabbitmq queue. I'm curious to know what others think.
2. About removing replicas if autoscaling is enabled, why do we need to change the current behaviour? Did you have any problem with it ?

pservit:
1. cpu autoscaling better than no autoscaling :)
2. during helm upgrade deployment scales down to number of replicas specified in deployment, than hpa need time to scale again, see https://github.com/kubernetes/kubernetes/issues/25238 for discussion

Mokto:
Ok, then maybe we could put a comment in the values.yaml file saying that it might not be the best metric to scale on ?

Thanks!

pservit:
Can you write a manual on how to setup autoscaling based on rabbitmq queue ?

Mokto:
That's a good idea. I'll merge your code and then write a quick guide on how to setup this.

PR #76, opened 2020-06-04
Fix templating error when disabling hooks

joscha-alisch:
Currently when disabling the hooks via 

```yaml
hooks:
  enabled: false
```

helm upgrade fails due to snuba-db-init and snuba-migrate still containing some yaml data. 

```yaml
---
# Source: sentry/templates/hooks/snuba-db-init.job.yaml
volumes:
        - name: config
          configMap:
            name: sentry-snuba
---
# Source: sentry/templates/hooks/snuba-migrate.job.yaml
volumes:
        - name: config
          configMap:
            name: sentry-snuba
```

This PR fixes that by actually removing the complete resource when setting enabled=false.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 1.3.0
+version: 1.3.1
 appVersion: 10.0.0
 dependencies:
   - name: redis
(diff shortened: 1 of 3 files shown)


Mokto:
Thanks! Could you update the patch version please ?

joscha-alisch:
Yep, sorry wasn't sure if you would want to do this :)

Mokto:
Thanks !!

PR #83, opened 2020-06-11
feat: added different types of snuba consumers

Mokto:
I'm still a bit unsure about the necessity of this

EDIT: seems like it consumed a lot of messages when it started, so it did something.


sentry/templates/deployment-snuba-outcomes-consumers.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: apps/v1
```
J0sh0nat0r: This file should probably be named `snuba-outcomes-consumer` to fit with the others

sentry/templates/deployment-snuba-outcomes-consumers.yaml
```diff
@@ -0,0 +1,71 @@
...
+        volumeMounts:
+        - mountPath: /etc/snuba
+          name: config
+          readOnly: true
+        resources:
+{{ toYaml .Values.snuba.outcomesConsumer.resources | indent 12 }}
+      volumes:
+        - name: config
+          configMap:
+            name: {{ template "sentry.fullname" . }}-snuba
```
J0sh0nat0r: Should probably end with a new line here

J0sh0nat0r (commented):
Other than the aforementioned LGTM 👍

sentry/templates/deployment-snuba-outcomes-consumers.yaml
Mokto: 👍

PR #88, opened 2020-06-18
Allow specifying tolerations on all clickhouse resources

joscha-alisch:
This allows specifying tolerations on all clickhouse related pods. 

I'm not sure how your process is for updating this dependency in the sentry chart, so I'll leave that to you :)


Mokto:
Thanks.
I'll take care of updating sentry chart now ;)

PR #90, opened 2020-06-19
Fix indentation issue

joscha-alisch:
Just noticed that my IDE did some auto-formatting when commiting, so my previous PR broke the YAML generation.. 

Sorry about that, it's fixed now.


Mokto:
Thanks

PR #91, opened 2020-06-20
Update Sentry & Snuba and add `migrations` dataset

J0sh0nat0r:
This PR adds the `migrations` dataset to maintain support for the latest tags and updates the Sentry & Snuba images accordingly. Unfortunately this introduces a backwards incompatability due to getsentry/snuba#1016 not being included in the current default tags, so I've had to bump the major chart version (again!).


Mokto:
Just tested it, it seems to work great!
Thanks.

And they finally switched to CalVer 👍

PR #104, opened 2020-06-28
Exposed clickhouse init container

ofekinger:
Moved busybox definitions from inside the template to values.yaml to enable image override and the use of private docker registries.


Mokto:
Lgtm ! Could you increase the minor version please?

Mokto (approved):

PR #66, opened 2020-05-11
Separate config for external services

donbeave:
Hello!

Thank you for this chart. I'm playing with it in the production environment. Until now everything works well.

I just found out some configuration confuses a little bit. There is best practice to use separate config for embedded charts (dependencies) and external services. It looks clear and predictable. That's why I did these changes in this PR. You can also check Bitnami repo (they actually one of the biggest helm charts contributor), they use the same way to split up config:
* wordpress chart: https://github.com/bitnami/charts/blob/8dce093c90ad398fa2a76a911a4ce95af3d4e5c0/bitnami/wordpress/values.yaml#L395
* kafka chart: https://github.com/bitnami/charts/blob/f45f62270a7b9e91c5dbedacdacfd12eb5b72398/bitnami/kafka/values.yaml#L823

Here is an example of values files:

```yaml
clickhouse:
  # disabled to use external clickhouse
  enabled: false

externalClickhouse:
  host: "clickhouse.clickhouse.svc.cluster.local"
  tcpPort: 9000

kafka:
  # disabled to use external kafka
  enabled: false

externalKafka:
  host: "kafka-myapp.kafka.svc.cluster.local"
  port: 9092

redis:
  # disabled to use external redis
  enabled: false

externalRedis:
  host: "redis-myapp-master.redis.svc.cluster.local"
  port: 6379

postgresql:
  # disabled to use external postgresql
  enabled: false

externalPostgresql:
  host: "postgresql-master.postgresql.svc.cluster.local"
  port: 5432
  username: sentry
  password: "my_sentry_password_here"
  database: sentry
```

PR #106, opened 2020-06-29
Enabled defining external kafka/zookeeper

jzuijlek:
Allow the user to define external Kafka and Zookeeper services like with PostgreSQL


jzuijlek:
https://github.com/sentry-kubernetes/charts/pull/66 is a better solution.

PR #85, opened 2020-06-11
Adding statsd backend + prometheus-operator support

ts-mini:
There are _some_ changes to the readme - i can either back these out or keep going. it just kinda is a pain :P - but metrics work for our installation and figured i'd share
Fixes https://github.com/sentry-kubernetes/charts/issues/81

PR #109, opened 2020-07-02
Add prometheus metrics to clickhouse

joscha-alisch:
This adds the option to enable the clickhouse prometheus exporter to run alongside clickhouse and create a servicemonitor resource for the prometheus operator.

Added Config: 
```yaml
clickhouse:
  metrics:
    prometheus:
      enabled: false
      exporter:
        repository: f1yegor/clickhouse-exporter
        tag: latest
        port: 9116
      serviceMonitor:
        enabled: false
```

In combination with #85 and the already existing support from the chart dependency, this would allow us to monitor every sentry component via prometheus.

(I didn't figure out how to update the chart.lock properly, so that is probably something you'd still need to do?)


Mokto:
Thanks! This PR should only contain changes related to clickhouse.

And when it is merged, the helm deployment will be made and I'll update the clickhouse dependency in sentry.

Hope I'm clear ;)

joscha-alisch:
Of course, I reverted the sentry changes :)

clickhouse/values.yaml
```diff
@@ -39,6 +39,21 @@ clickhouse:
   ##
   # rollingUpdatePartition:
 
+  metrics:
+    prometheus:
+      ###
+      # Enabling this starts a clickhouse prometheus exporter as an extra container in all clichouse pods and exposes it in the service under port "metrics"
```
DandyDeveloper: *Clickhouse 

Long comment, can you throw it on 2 lines?

DandyDeveloper (changes requested):
I'd highly recommend adding and differentiating between Prometheus operator and Prometheus itself. 

Both work differently for target discovery. 

Prometheus (non-operator) will rely on service annotations and/or pod annotation for scraping. So we'd need to support this. Additionally, you'd need to add the port to the clickhouse `Service`, as opposed to a new `ServiceMonitor`. 

You current config is more for a Prometheus operator. 

We should probably support both use-cases and make the distinction clear in the values file.

joscha-alisch:
That makes sense to me @DandyDeveloper. 
However, I only use prometheus-operator myself, so I am not entirely sure what needs to be added to support a standalone prometheus?

DandyDeveloper:
> 
> 
> That makes sense to me @DandyDeveloper.
> However, I only use prometheus-operator myself, so I am not entirely sure what needs to be added to support a standalone prometheus?

@joscha-alisch 
This is one of the easier to follow resources. It's effectively adding the port to the service AND annotations.
https://www.weave.works/docs/cloud/latest/tasks/monitor/configuration-k8s/

joscha-alisch:
Ah great, I'll look into that.

In the meantime, I found the following values.yaml of the [bitnami redis chart](https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml#L556-L652), which appears like a good structure to me. What do you think? I'd model it according to that:


```yaml
metrics:
  enabled: false

  image:
    registry: docker.io
    repository: bitnami/redis-exporter
    tag: 1.8.0-debian-10-r12
    pullPolicy: IfNotPresent

  ## Metrics exporter resource requests and limits
  resources: {}

  ## Metrics exporter pod Annotation and Labels
  podAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "9121"
   podLabels: {}

  # Enable this if you're using https://github.com/coreos/prometheus-operator
  serviceMonitor:
    enabled: false
    ## Specify a namespace if needed
    # namespace: monitoring
    # fallback to the prometheus default unless specified
    # interval: 10s
    ## Defaults to what's used if you follow CoreOS [Prometheus Install Instructions](https://github.com/bitnami/charts/tree/master/bitnami/prometheus-operator#tldr)
    ## [Prometheus Selector Label](https://github.com/bitnami/charts/tree/master/bitnami/prometheus-operator#prometheus-operator-1)
    ## [Kube Prometheus Selector Label](https://github.com/bitnami/charts/tree/master/bitnami/prometheus-operator#exporters)
    selector:
      prometheus: kube-prometheus

  ## Metrics exporter pod priorityClassName
  # priorityClassName: {}
  service:
    type: ClusterIP
    ## Use serviceLoadBalancerIP to request a specific static IP,
    ## otherwise leave blank
    # loadBalancerIP:
    annotations: {}
    labels: {}

joscha-alisch:
@DandyDeveloper , I pushed the changes. The metrics are now configurable just like the above mentioned bitnami chart. I hope that better fits what you had in mind. Let me know if you need me to change anything about this :)

DandyDeveloper (approved):
LGTM (Sorry about the delays)

DandyDeveloper:
@Mokto This is good to merge whenever you feel comfortable

Mokto:
Thanks guys 👍

PR #123, opened 2020-07-21
sentry: remove the default secret-key

yoan-adfinis:
Providing a default secret key is probably a security issue if it doesn't get changed.


Mokto:
Ah, thanks!

I'm not sure about this one to be honest. I like the simplicity about installing it without anything required (at least for testing purposes).

Maybe just adding a note to the readme would be enough?

@DandyDeveloper @J0sh0nat0r what do you think?

yoan-adfinis:
When _testing_ purposes end up on the internet people are crying.

DandyDeveloper:
@yoan-adfinis @Mokto Both implementations are wrong IMO, although always up for debate.

We should leave nil and generate a random pass, then in the output of Helm, pass this to the user: 

`{{ default (randAlphaNum 15) .Values.somePassword | b64enc | quote }}`

This does mean it's regenerated on upgrade, so we need to make sure it's documented that this password needs to be passed next time. 

This is common practice for most mature charts.

sentry/values.yaml
```diff
@@ -148,7 +148,7 @@ hooks:
 system:
   url: ""
   adminEmail: ""
-  secretKey: 'icLq77rCyY_qrMMpXa6TQNjkDV6mU!c'
+  secretKey: ~  # This will generate one for you (it's must be given upon updates)
```
Mokto: I would assume we need to comment this line instead ?

sentry/README.md
```diff
@@ -55,3 +55,4 @@ Parameter                          | Description
 `metrics.serviceMonitor.honorLabels` | honorLabels chooses the metric's labels on collisions with target labels. | `false`
 `metrics.serviceMonitor.namespace` | namespace where servicemonitor resource should be created | `the same namespace as sentry`
 `metrics.serviceMonitor.scrapeInterval` | interval between Prometheus scraping | `30s`
+`sentry.secretKey` | **Required** secret key for the session cookie ([documentation](https://develop.sentry.dev/config/#general)) | `nil`
```
Mokto: I think you can remove *required*.

Also, could you add a specific note to the readme ?
Similar to https://github.com/sentry-kubernetes/charts#postgressql

Mokto (changes requested):

Mokto:
Agreed @DandyDeveloper.

And thanks for your work on this @yoan-adfinis

Mokto:
Thanks!

PR #125, opened 2020-07-22
Add legacyApp field for slack

mrtc0:
For the integration with the Slack App to work, need to configure `slack.legacy-app` .

ref : https://github.com/getsentry/sentry/pull/19446

### DIFF
--- a/sentry/templates/configmap-sentry.yaml
+++ b/sentry/templates/configmap-sentry.yaml
@@ -49,7 +49,7 @@ data:
     slack.client-secret: "{{ .Values.slack.clientSecret }}"
     slack.verification-token: "{{ .Values.slack.verificationToken }}"
     {{- if .Values.slack.legacyApp }}
-    slack.legacy-app: "{{ .Values.slack.legacyApp }}"
+    slack.legacy-app: {{ .Values.slack.legacyApp }}
     {{ end }}
     {{ end }}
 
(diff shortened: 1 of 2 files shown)


sentry/templates/configmap-sentry.yaml
```diff
@@ -48,6 +48,9 @@ data:
     slack.client-id: "{{ .Values.slack.clientId }}"
     slack.client-secret: "{{ .Values.slack.clientSecret }}"
     slack.verification-token: "{{ .Values.slack.verificationToken }}"
+    {{- if .Values.slack.legacyApp }}
+    slack.legacy-app: "{{ .Values.slack.legacyApp }}"
```
Mokto: Thanks !
From what I see it should be a "python" boolean, not a string. Have you tried the behaviour on your instance?

sentry/values.yaml
```diff
@@ -192,6 +192,7 @@ slack: {}
 #   clientId:
 #   clientSecret:
 #   verificationToken:
+#   legacyApp:
```
Mokto, line 199: Should we put another comment to reference this issue ?
https://github.com/getsentry/sentry/pull/19446

sentry/templates/configmap-sentry.yaml
mrtc0: I haven't tried it in an instance. You're right, it should be a boolean here. I'll fix it.

sentry/values.yaml
mrtc0: It's certainly better to be able to reference the issue.

sentry/templates/configmap-sentry.yaml
Mokto: If you put `#   legacyApp: true` in the values
and then

```

 {{- if .Values.slack.legacyApp }}
    slack.legacy-app: True
{{- else }}
    slack.legacy-app: False
```

That would make it easy to use

mrtc0: Thanks for the good suggestions. fixed it.

Mokto:
Thanks!

PR #130, opened 2020-07-22
feat: relay support

### LINKED ISSUE #86: Add sentry relay
https://github.com/getsentry/relay

### LINKED ISSUE #127: CSRF Token issue. No events received with 20.7.2 / 20.7.1
See https://forum.sentry.io/t/csrf-token-issue/9607/3.

We probably have to introduce nginx + relay, as in the on promise installation.

Mokto:
Fix #86 
Fix #127

### DIFF
--- a/sentry/templates/deployment-nginx.yaml
+++ b/sentry/templates/deployment-nginx.yaml
@@ -1,0 +1,92 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "sentry.fullname" . }}-nginx
+  labels:
+    app: {{ template "sentry.fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  selector:
+    matchLabels:
+        app: {{ template "sentry.fullname" . }}
+        release: "{{ .Release.Name }}"
+        role: nginx
+{{- if not .Values.nginx.autoscaling.enabled }}
+  replicas: {{ .Values.nginx.replicas }}
+{{- end }}
+  template:
+    metadata:
+      annotations:
+        checksum/config.yaml: {{ include (print $.Template.BasePath "/configmap-nginx.yaml") . | sha256sum }}
+        {{- if .Values.nginx.annotations }}
+{{ toYaml .Values.nginx.annotations | indent 8 }}
+        {{- end }}
+      labels:
+        app: {{ template "sentry.fullname" . }}
+        release: "{{ .Release.Name }}"
+        role: nginx
+        {{- if .Values.nginx.podLabels }}
+{{ toYaml .Values.nginx.podLabels | indent 8 }}
+        {{- end }}
+    spec:
+      affinity:
+      {{- if .Values.nginx.affinity }}
+{{ toYaml .Values.nginx.affinity | indent 8 }}
+      {{- end }}
+      {{- if .Values.nginx.nodeSelector }}
+      nodeSelector:
+{{ toYaml .Values.nginx.nodeSelector | indent 8 }}
+      {{- end }}
+      {{- if .Values.nginx.tolerations }}
+      tolerations:
+{{ toYaml .Values.nginx.tolerations | indent 8 }}
+      {{- end }}
+      {{- if .Values.images.nginx.imagePullSecrets }}
+      imagePullSecrets:
+{{ toYaml .Values.images.nginx.imagePullSecrets | indent 8 }}
+      {{- end }}
+      containers:
+      - name: {{ .Chart.Name }}-nginx
+        image: "{{ .Values.images.nginx.repository }}:{{ .Values.images.nginx.tag }}"
+        imagePullPolicy: {{ .Values.images.nginx.pullPolicy }}
+        # command: ["sentry", "run", "nginx"]
+        ports:
+        - containerPort: {{ template "nginx.port" }}
+{{- if .Values.nginx.env }}
+{{ toYaml .Values.nginx.env | indent 8 }}
+{{- end }}
+        volumeMounts:
+        - mountPath: /work/.nginx
+          name: config
+          readOnly: true
+        livenessProbe:
+          failureThreshold: 5
+          httpGet:
+            path: /_health/
+            port: {{ template "nginx.port" }}
+            scheme: HTTP
+          initialDelaySeconds: {{ .Values.nginx.probeInitialDelaySeconds }}
+          periodSeconds: 10
+          successThreshold: 1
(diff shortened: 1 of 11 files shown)


walkerdb:
hey @Mokto anything blocking this from release? Happy to help if there's anything you need.

Mokto:
Sorry, i have been really busy those past few days.

I have had issues creating the credentials.json file properly.

Mokto:
If you want to take over, feel free to do it! I'm also leaving for vacations soon, so I probably won't have time to anything before 3 weeks.

Thanks.

chhetripradeep:
Just want to add one thing:

We need to explicitly add an env variable RELAY_PORT with value 3000 in relay deployment. I ran into an issue due to that. For detailed info: https://github.com/getsentry/relay/issues/672

Thanks !!

walkerdb:
poked at this briefly and yeah generating credentials is going to be a problem.

A brief summary:
* The relay service requires two config files to run: one normal `config.json` with service configuration, and a `credentials.json` file that contains an id and a public/private key pair [generated with `ed25519`](https://github.com/getsentry/relay/blob/master/relay-auth/src/lib.rs#L420).
* we can hardcode the config.json contents, but the credentials must manually be generated before we can start the service up. Normally the credentials are generated using the relay cli, which itself needs the `config.json` file to run, eg as `docker run -v $(pwd)/config.yml:/tmp/config.yml getsentry/relay:latest --config /tmp credentials generate --stdout`
* helm can't run arbitrary commands before starting a chart, and we likely don't want to hardcode a credentials file with literal secrets in it

this all means we'll need to require the consumer of this helm chart to supply those credentials somehow. Some initial thoughts on how to go about it:

* expose template values for each `credentials.json` field, and fill them into a relay configmap entry for `credentials.json`. The consumer would need to run the relay credentials command on their own to get these values.
* require that the consumer create a literal `credentials.json` file on their own that will be referenced by the chart, and added as a volume to the relay image when it starts up
* require consumers to create their own relay images with config and credentials already set (not a fan of this one)

All of those options make the chart a little less "just run helm and go" since they add some required pre-work before anything can run, but I'm not sure if there's any better option. Any thoughts on this one?

sidenote: sentry also needs to be configured to accept the public key generated for relay, but I'm not finding any obvious way to do this purely with config, since the sentry docs only reference ways to do this from the GUI and I'm not seeing any obvious code paths for it. If we can't automate this then any new instance using this chart won't work until the user manually goes into the sentry UI to add in the public key. I'm hoping there's a config path for this that I just haven't seen yet.

chhetripradeep:
Hi @walkerdb 

Thank you for sharing your learnings. 

Regarding the credential automatic generation, we can make use of init container (https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) with init container sharing the filesystem with main relay container and run command `relay credentials generate` as part of init container. I am using something like this.

```
      initContainers:
        - name: {{ .Values.relay.name }}-init
          image: {{ .Values.relay.image.repository }}:{{ .Values.relay.image.tag }}
          args:
            - "credentials"
            - "generate"
          env:
            - name: RELAY_PORT
              value: "{{ .Values.relay.port }}"
          volumeMounts:
            - name: credentials
              mountPath: /work/.relay
            - name: config
              mountPath: /work/.relay/config.yml
              subPath: config.yml
              readOnly: true
      volumes:
        - name: config
          configMap:
            name: {{ .Values.relay.name }}
            defaultMode: 0644
        - name: credentials
          emptyDir: {}
```

Although the relay docs says to manually register each relay process using its public key, from my experience, they get auto-registered as they start without any manual step requirement.

Relay logs 
```
2020-08-02T12:50:24Z [relay_server::actors::upstream] INFO: upstream relay started
2020-08-02T12:50:24Z [relay_server::actors::upstream] INFO: registering with upstream (http://sentry-web:9000/)
```

sentry/templates/deployment-nginx.yaml
```diff
@@ -0,0 +1,92 @@
...
+{{ toYaml .Values.nginx.env | indent 8 }}
+{{- end }}
+        volumeMounts:
+        - mountPath: /etc/nginx
+          name: config
+          readOnly: true
+        livenessProbe:
+          failureThreshold: 5
+          httpGet:
+            path: /_health/
```
chhetripradeep, line 67: The healtcheck endpoint is wrong: it should be `/api/relay/healthcheck/ready/` and same for readinessProbe too.

sentry/templates/deployment-relay.yaml
```diff
@@ -0,0 +1,93 @@
...
+{{ toYaml .Values.relay.env | indent 8 }}
+{{- end }}
+        volumeMounts:
+        - mountPath: /work/.relay
+          name: config
+          readOnly: true
+        livenessProbe:
+          failureThreshold: 5
+          httpGet:
+            path: /_health/
```
chhetripradeep: The health check endpoint is wrong: it should be `/api/relay/healthcheck/ready/` and same for readinessProbe too.

sentry/templates/deployment-nginx.yaml
chhetripradeep: Would suggest to use nginx specific health check endpoint here.

sentry/templates/deployment-relay.yaml
Mokto: Thanks! Was not really ready for review, but that'll help :)

sentry/templates/deployment-nginx.yaml
Mokto: Good point. I changed it.

Mokto:
@chhetripradeep Thanks for your help.

I'm now stuck at the step where we need to register the credentials in sentry, I suppose:

```
2020-08-03T07:55:33Z [relay_server::actors::upstream] ERROR: authentication encountered error: could not send request to upstream
  caused by: Server disconnected
  caused by: Server disconnected
```

chhetripradeep:
Hi @Mokto 

Sorry I am not sure about it. I am also in process of migrating to relay based setup and things aren't working for me too. I will try out and let you know. Thank you.

yoan-adfinis:
Would it make sense to require an `ingress-controller` to the setup rather than embedding nginx?

Mokto:
I think the default should be with Nginx! (easier to setup).
But it would be nice to (optionally) allow ingress-controller (or istio/linkerd) to prevent using nginx

yoan-adfinis:
@Mokto It did try something on #137. However it seems to fail when it comes to having the event into kafka :shrug:

Mokto:
Thanks! 
So you basically fixed the credentials issue by applying 
`mode: managed`?

yoan-adfinis:
afaik, the _managed_ mode is the default one. https://getsentry.github.io/relay/configuration/modes/

The _redis_ URLs looked strange however.

PR #136, opened 2020-07-31
sentry: fix appVersion

yoan-adfinis:
version has been reverted in https://github.com/sentry-kubernetes/charts/pull/128 but not the `appVersion`


Mokto:
Thanks.

PR #140, opened 2020-08-06
Add configuration option for commit batch size

### LINKED ISSUE #138: Post process forwarder: commit-batch-size=1
Hi there :) 

is there specific reason why commit batch size is set to 1 in the post-process-forward deployment? I've seen the app be able to process a much higher volume of post processing when I remove this option. 

In case this can be removed, I am happy to provide a PR, either removing it altogether or at least making it configurable :)

joscha-alisch:
Fixes #138


Mokto:
Thanks!

PR #137, opened 2020-08-04
Feat/sentry with relay

yoan-adfinis:
An alternative to #130 without the nginx requirement. In my experiments, you must disable the relay for the db init to occur, then scale it up. There is probably a nice way to do so via Helm but I ignore it.

I've managed to bypass the CSRF problem but not yet the event.

```
2020-08-04T06:26:48Z [relay_server::actors::project_upstream] DEBUG: updating project states for 1/1 projects (attempt 1)
2020-08-04T06:26:48Z [relay_server::actors::project_upstream] DEBUG: sending request of size 1
2020-08-04T06:26:48Z [relay_server::actors::project] DEBUG: project 2 state updated
2020-08-04T06:26:48Z [relay_server::actors::events] DEBUG: dropped event: could not send event to upstream
  caused by: could not send request to upstream
  caused by: Server disconnected
  caused by: Server disconnected
```

PR #141, opened 2020-08-07
feat(sentry): use AppVersion as the default image tag

yoan-adfinis:
Make it easier to avoid out of sync versions in the future (with 20.7.2 using `AppVersion` should just be fine, #137)

### DIFF
--- a/sentry/templates/_helper.tpl
+++ b/sentry/templates/_helper.tpl
@@ -12,6 +12,22 @@
 {{- define "sentry.port" -}}9000{{- end -}}
 {{- define "snuba.port" -}}1218{{- end -}}
 
+{{- define "relay.image" -}}
+{{- default "getsentry/relay" .Values.images.relay.repository -}}
+:
+{{- default .Chart.AppVersion .Values.images.relay.tag -}}
+{{- end -}}
+{{- define "sentry.image" -}}
+{{- default "getsentry/sentry" .Values.images.sentry.repository -}}
+:
+{{- default .Chart.AppVersion .Values.images.sentry.tag -}}
+{{- end -}}
+{{- define "snuba.image" -}}
+{{- default "getsentry/snuba" .Values.images.snuba.repository -}}
+:
+{{- default .Chart.AppVersion .Values.images.snuba.tag -}}
+{{- end -}}
+
 {{/*
 Expand the name of the chart.
 */}}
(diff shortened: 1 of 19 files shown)


sentry/templates/_helper.tpl
```diff
@@ -12,6 +12,22 @@
...
+{{- end -}}
+{{- define "sentry.image" -}}
+{{- default "getsentry/sentry" .Values.images.sentry.repository -}}
+:
+{{- default .Chart.AppVersion .Values.images.sentry.tag -}}
+{{- end -}}
+{{- define "snuba.image" -}}
+{{- default "getsentry/snuba" .Values.images.snuba.repository -}}
+:
+{{- default .Chart.AppVersion .Values.images.snuba.tag -}}
```
DandyDeveloper, line 28: Shouldn't this be a `tpl` with the context passed in? `tpl .Chart.AppVersion .`?

DandyDeveloper (commented):
I'd need to triple check, but doesn't refs like this need to be a tpl in the templating?

sentry/templates/_helper.tpl
yoan-adfinis: It seems to work as is.

```console
$ helm install --dry-run --generate-name --debug . | grep "getsentry\\/"
install.go:159: [debug] Original chart version: ""
install.go:176: [debug] CHART PATH: /home/yoan/soft/charts/sentry

        image: "getsentry/sentry:20.7.2"
        image: "getsentry/sentry:20.7.2"
          image: "getsentry/relay:20.7.2"
        image: "getsentry/relay:20.7.2"
        image: "getsentry/sentry:20.7.2"
        image: "getsentry/sentry:20.7.2"
        image: "getsentry/sentry:20.7.2"
        image: "getsentry/sentry:20.7.2"
        image: "getsentry/snuba:20.7.2"
        image: "getsentry/snuba:20.7.2"
        image: "getsentry/snuba:20.7.2"
        image: "getsentry/snuba:20.7.2"
```

DandyDeveloper: Interesting! I also gave bad code haha, but good to know.

Mokto:
Thanks! I think this is ready for merging, right?

hairmare:
> Thanks! I think this is ready for merging, right?

I just added the template calls to all the new deployments and hooks from 5.1.0, so yeah I think so.

Mokto:
Ah, good catch! Merging now

PR #144, opened 2020-08-10
Feat/sentry ingest consumer

yoan-adfinis:
This pull request enables relay and the ingest-consumer on the current version of Sentry. A follow up of #137 (and #130)

**NB** this seems to work on a fresh install but I've seen some 403 when doing an upgrade from the current version. :confused:


yoan-adfinis:
@Mokto this version includes the `nginx`  for internal cluster usage I guess. I'm assuming many things are missing if this was to be used within the cluster.

Mokto:
Lgtm ! 
I would add some notes in the readme to explain why we need nginx (explaining that we should either use nginx, either the ingress).

Thanks for your work on this ;)

Mokto:
& change the version to 5.0.0 ;)

Mokto:
@yoan-adfinis 
I'm taking the liberty to add a commit to your branch. Hope that's fine?
I'm leaving for vacations tomorrow (2 weeks), so I'd like this PR to be merged today ;) 

Redis connection is not working with external redis (probably it's the same for kafka, I'll try)

yoan-adfinis:
> Hope that's fine?

Absolutely.

> I'm leaving for vacations tomorrow

Same same, kind of. I already wish you some great time off.

Mokto:
@yoan-adfinis can you check my commit ?
I think everything is working fine now :)

Enjoy your vacations too!

PR #148, opened 2020-08-13
sentry: Use existingSecret for psql

davelosert:
I realized that the sentry-chart currently uses a static values for the postgresql-secret.

This is a bit counter-intuitive and will fail when using the `existingSecret` and `existingSecretKey`-Option provided by the [bitnami-postgresql helm chart](https://hub.helm.sh/charts/bitnami/postgresql).

This MR uses those secrets if they are set and falls back to the current values if not.


Mokto (approved):

Mokto:
Thanks!

Issue #153, opened 2020-08-18
django.security.csrf: Forbidden


ivanblack19:
After upgrade to chart 5.0.0
We can see in sentry-web logs:

>  [WARNING] django.security.csrf: Forbidden (Referer checking failed - no Referer.): /api/2/store/ (status_code=403 request=<WSGIRequest: POST u'/api/2/store/'>)
> [18/Aug/2020:13:43:25 +0000] "POST /api/2/store/ HTTP/1.1" 403 5667 "-" "Go-http-client/2.0"

Logs from backend NOT ingested,
We DO NOT use NGINX or INGRESS (we use Google Load Balancer) to access Sentry.

part of config as follows

```
nginx:
  enabled: false
  containerPort: 8080
  existingServerBlockConfigmap: sentry-nginx
  resources: {}
  replicaCount: 1
  service:
    type: ClusterIP
    port: 80

ingress:
  enabled: false
  # annotations:
  #   kubernetes.io/tls-acme:
  #   certmanager.k8s.io/issuer:
  #
  #   nginx.ingress.kubernetes.io/proxy-body-size:
  #   nginx.ingress.kubernetes.io/use-regex: "true"
  #
  #   nginx.ingress.kubernetes.io/enable-cors: "true"
  #   nginx.ingress.kubernetes.io/cors-allow-methods: "PUT,GET,POST,OPTIONS"
  #   nginx.ingress.kubernetes.io/cors-allow-origin: "*"
  #   nginx.ingress.kubernetes.io/cors-allow-credentials: "true"
  #
  # hostname:
  #
  # tls:
  # - secretName:
  #   hosts:
```

Mokto:
Hey,

Can you share your Google Load Balancer config somehow?

You have to redirect the store events on relay. If you don't, you get csrf issues.

ivanblack19:
Could you suggest how to redirect the store events on relay ?

As for Google Load Balancer we use a standard NEG (network endpoint group) in Sentry Service 

```
service:
  type: NodePort
  annotations:
    cloud.google.com/neg: '{"exposed_ports": {"9000":{}}}'
```

And then we create standard HTTP(S) Global Load balancer in Google

<img width="1015" alt="image" src="https://user-images.githubusercontent.com/56841680/91417729-e6cab880-e859-11ea-8fb7-46227ebeb7df.png">

Mokto:
Ah, ok! That's why it doesn't work.

You can either :
[...]

PR #155, opened 2020-08-20
fix: django.security.csrf issue

tetsun:
https://github.com/sentry-kubernetes/charts/issues/153
I got the same problem.

Nginx should set a host header for csrf.

https://github.com/getsentry/onpremise/pull/463
This PR seems to solve this problem. (It works in my k8s)


Mokto:
Thanks! I was indeed an issue for new logins.

PR #158, opened 2020-08-24
sentry: relay mode config

GermanGlushkov:
Relay pod is crashing in current version of chart (5.0.0), changing relay mode to "proxy" helps. Default "managed" mode doesn't seem to be supported https://getsentry.github.io/relay/

### DIFF
--- a/sentry/templates/configmap-relay.yaml
+++ b/sentry/templates/configmap-relay.yaml
@@ -13,7 +13,7 @@ metadata:
 data:
   config.yml: |-
     relay:
-      mode: managed
+      mode: {{ .Values.relay.mode }}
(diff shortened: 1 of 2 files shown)


Mokto:
Does anyone else have this issue ? 
The managed mode seems to work fine for me.

I would say let's merge your PR but let's set "managed" as default in the value.yaml file.

J0sh0nat0r:
Haven't seen any issues here, would you mind sharing the pod logs?

GermanGlushkov:
You're right, the root cause seems to be istio, my apologies. I've changed default mode to "managed", I can as well close this PR if you think this change is not actually needed.

Mokto:
Cool you found the issue.

I think we can close this for now, thanks anyway!

PR #165, opened 2020-08-30
Special characters in email and password

achichenin:
Currently, this job silently fails if special characters due to missing quoting. Straightforward fix.


Mokto:
Thanks!

Issue #84, opened 2020-06-11
Add cleanup jobs with a CronJob


Mokto:
Should run:
- Sentry
- Snuba
- Symbolicator

DandyDeveloper:
@Mokto I wonder if this could replace the cron pod entirely? I still don't fully understand what the cron pod does...

Mokto:
The weekly report for example :

https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L673

jmijn:
Also missing a configuration option in the values to set the retention days. Now its set to 90 always:
https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/configmap-sentry.yaml#L138

hairmare:
👍 on getting rid of the cron pod entirely! I didn't take the time to figure out what it does but I'd totally agree that all of it should be achievable through proper K8s CronJobs.

BYK:
The `cron` worker on Sentry would still be needed as it is actually celery-beat but other `-cleanup` images built from base images with Cron layered on top should be trivial to get rid of with k8s scheduled event runs. The reason we have them this way is `docker-compose` not having this functionality.

Mokto:
Closing as 5.2.0 solves this issue.

PR #178, opened 2020-09-21
Add sentry cleanup cronjob

justincmoy:
https://github.com/sentry-kubernetes/charts/issues/84: Add cleanup jobs with a CronJob

Defaults taken from https://github.com/getsentry/onpremise/blob/21d1b6d42eee14d0de93ba77fe981f9f2a82447e/docker-compose.yml#L184-L191 and https://github.com/getsentry/onpremise/blob/21d1b6d42eee14d0de93ba77fe981f9f2a82447e/.env#L2.

PodSpec inspired by the cron deployment: https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/deployment-sentry-cron.yaml

### DIFF
--- a/sentry/values.yaml
+++ b/sentry/values.yaml
@@ -108,6 +108,10 @@ sentry:
     # tolerations: []
     # podLabels: []
     # commitBatchSize: 1
+  cleanup:
+    enabled: false
+    schedule: "0 0 * * *"
+    days: 90
 
 snuba:
   api:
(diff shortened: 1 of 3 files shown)


sentry/values.yaml
```diff
@@ -108,6 +108,10 @@ sentry:
     # tolerations: []
     # podLabels: []
     # commitBatchSize: 1
+  cleanup:
+    enabled: false
```
Mokto: Thanks for your PR.

I think we should enable this per default ?

justincmoy: Done!

Mokto:
Checked and works fine :)

Thanks for your PR!

PR #179, opened 2020-09-22
fix: relay deployment

Mokto:


sentry/templates/deployment-relay.yaml
```diff
@@ -7,6 +7,10 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  annotations:
+    "helm.sh/hook": "post-install,post-upgrade"
+    "helm.sh/hook-weight": "15"
+    "helm.sh/resource-policy": "keep"
```
J0sh0nat0r: Won't this prevent relay from getting cleaned up on uninstall, thus breaking upgrades?
Going off of the docs [here](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) although I could be going wrong if it's different for hooks

Mokto: Good point, I'll try tomorrow with both.

DandyDeveloper: :+1: Nothing else I can see is wrong with this.

PR #180, opened 2020-09-22
chore: added a note for nginx ingress controller

Mokto:


DandyDeveloper (commented):
Could we enabled enable these automatically if `xyz.ingress.enabled` = True?

odoucet:
> Could we enabled enable these automatically if `xyz.ingress.enabled` = True?

Nope, because nginx may not be the default ingress in the k8s cluster. 

Adding comments si good enough and will help user :)

DandyDeveloper:
@odoucet Sure, but you could look at the annotation value... There's a lot of nice ways you could achieve this.

Mokto:
I'll merge it like that for now, but it can definitely be improved.

PR #182, opened 2020-09-24
feat: sentry 20.9.0

Mokto:


J0sh0nat0r (approved):
LGTM 👍  The level of sync here pleases my OCD 😆

PR #185, opened 2020-09-24
fix: nginx configmap name in values.yaml

slumbi:
The nginx configmap name was hardwired to `sentry-nginx`, and it only works if the helm release name is `sentry`.
This commit fixes that by using a templated variable instead


Mokto:
Thanks ! i didn't know about templated variables

Issue #134, opened 2020-07-27
duplicated nodePort declaration in template 


ruseel:
two file has same `nodePort` declaration

> sentry/templates/service-snuba.yaml
> sentry/templates/service-sentry.yaml

like 

> {{- if and (.Values.service.nodePort) (eq .Values.service.type "NodePort") }}
>     nodePort: {{ .Values.service.nodePort }}
> {{- end }}

so  If I activate this if statement 
I got error

> spec.ports[0].nodePort: Invalid value: 30001: provided port is already allocated

because two k8s `service` has save `nodePort`.

Mokto:
Yes, I also noticed recently that for services, every service is using that `service` value.

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

zemkogabor:
Is there any change regarding this? If I change the node port, the two services still conflict with each other.

tanyaofei:
Well, if you want to export a port, you should config it for nginx `nginx.service.nodePorts.http: 30991` and `nginx.service.type: NodePort`

PR #186, opened 2020-10-02
Updating Sentry config.yaml to support actual yaml.

DandyDeveloper:
I honestly am not sure if we're too far gone to actually do this... But we really should support YAML in the `configYml` option. 

Make pipelining and stuff MUCH easier using this.

### DIFF
--- a/sentry/templates/configmap-sentry.yaml
+++ b/sentry/templates/configmap-sentry.yaml
@@ -109,7 +109,9 @@ data:
       {{- end }}
     {{ end }}
 
-{{ .Values.config.configYml | indent 4 }}
+    {{- if .Values.config.configYml }}
+{{ .Values.config.configYml | toYaml | indent 4 }}
+    {{- end }}
   sentry.conf.py: |-
     from sentry.conf.server import *  # NOQA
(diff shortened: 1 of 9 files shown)


Mokto:
Yeah, the idea was to maintain compatibility with the old chart.

Now it's not the case anymore, but I'm a bit reluctant to change a major version just for that.

Maybe we should use this opportunity to separate the 4 services definitions at the same time ?

DandyDeveloper:
> Maybe we should use this opportunity to separate the 4 services definitions at the same time ?

@Mokto Can you expand on this

Mokto:
Talking about this issue : https://github.com/sentry-kubernetes/charts/issues/134

DandyDeveloper:
@Mokto I definitely think we should consider making "Larger" changes and noting the breaking changes in release notes.

This change for example is just very appropriate to me.

DandyDeveloper:
@Mokto Do we wanna do this?

Mokto:
Yes, I'm going to merge it and deploy 6.0.0 with https://github.com/sentry-kubernetes/charts/pull/198 which is also breaking.

PR #198, opened 2020-10-08
Allow some snuba deployments to be pushed as hooks

mariusmitrofanbostontr:


mariusmitrofanbostontr:
@Mokto do you mind if you push this change too please?

I hate having to scale down / up replicas.

The default functionality is to leave it "as it was" anyway :)

Mokto:
I'm thinking maybe we should combine asHook with the relay one.

So having one "asHook" for everything.

What do you think ?

mariusmitrofanbostontr:
Sounds good to me :)

Updated!

Mokto:
Thanks ! Could you remove the version bump?
I'm going to apply another fix + change the major version!

mariusmitrofanbostontr:
Sure, done

PR #203, opened 2020-10-11
fix: erroneous formating in relay configmap

### LINKED ISSUE #202: [5.6.0] Sentry relay crashing with default values.yaml
When trying to install the sentry chart, the relay pod always crashes. The default values have not been touched except for these fields:

```
user.email: "<hidden-email>"
user.password "<hidden-password>"

mail.backend: smtp
mail.useTls: true
mail.username: "<hidden>"
mail.password: "<hidden>"
mail.port: 465
mail.host: smtp.gmail.com
mail.from: "<hidden>"

slack.clientId: <hidden>
slack.clientSecret: <hidden>
slack.verificationToken: <hidden>
```

I can confirm that these values have worked before with the 5.2.0 version of the chart.
When trying to get some log output from the relay pod, the k8s server only responds with a BadRequest - even with the 
`--previous` flag

When looking at the pod through `kubectl describe` I also can't find any hints as to what's going on. The pod just keeps crashing out of nowhere.

Can somebody confirm this behaviour, since I'm not all to sure that it's not to do with something in my cluster configuration?
Note: The Chart is installed via a wrapper Helm Chart in a `sentry` namespace.

dnlsndr:
Due to trailing minuses in the conditionals, the yaml is geneated in an erroneous way such that there are no newlines before and after the relay.mode yaml key.
This leads to invalid yaml syntax causing the relay pod to crash.

fixes: #202
Signed-off-by: Daniel Schneider <daniel.schneider@eramux.com>

### DIFF
--- a/sentry/templates/configmap-relay.yaml
+++ b/sentry/templates/configmap-relay.yaml
@@ -13,9 +13,9 @@ metadata:
 data:
   config.yml: |-
     relay:
-      {{- if .Values.relay.mode -}}
+      {{- if .Values.relay.mode }}
       mode: {{ .Values.relay.mode }}
-      {{- end -}}
+      {{- end }}
       upstream: "http://{{ template "sentry.fullname" . }}-web:{{ .Values.service.externalPort }}/"
       host: 0.0.0.0
       port: {{ template "relay.port" }}


Mokto:
Thanks!

Issue #208, opened 2020-10-14
DSN endpoint changed and no option to go back to previous one


Leen15:
We just upgraded the chart to 6.0.0, and all ours DSN endpoints changed from:
`https://<sentry-dsn>@<subdomain>.<domain>.tld`
To:
`https://<sentry-dsn>@o<number>.ingest.<subdomain>.<domain>.<tld>`
We have a wildcard SSL that doesn't support oX.ingest.
We checked that this new option is available only if `organizations:org-subdomains` is set to True.
Unfortunately, there is no way on the chart to set it, because it's forced to true in the configmap:
https://github.com/sentry-kubernetes/charts/blob/945e4c0e44facbeae402fbb5f32a7ce05dd53dfe/sentry/templates/configmap-sentry.yaml#L327

Is there a way for set this option to false without manually change the configmap at every chart update?

Thanks.

mthoretton:
I'm also facing this problem.
Even if we wanted to use the `o<number>.ingest.<subdomain>.<domain>.<tld>` format, how would that work?
Would we have to manually update the sentry ingress to add a new domain any time there is a new organisation? (we use external-dns/ingress-nginx/cert-mananager)
I would also prefer to keep the good old endpoint.
Thanks 😃

Mokto:
I guess we can remove that value entirely ? But it probably means a new major release.

What do you think ?

Leen15:
@Mokto I agree, it should be removed with the ability to set in values file.

mthoretton:
I created https://github.com/sentry-kubernetes/charts/pull/209 which should solve this specific problem wihout introducing any breaking change. At least that would do the job for me 😄

mthoretton:
Thanks a lot for your reactivity. I was able to deploy the new chart version and I can confirm that it has fixed the issue (by setting sentry.features.orgSubdomains to `false`. 🎉

PR #209, opened 2020-10-14
Make organizations:org-subdomains feature optional

mthoretton:
See https://github.com/sentry-kubernetes/charts/issues/208 (DSN endpoint changed)

This PR allows `organizations:org-subdomains` sentry feature to be enabled or not.
Having different subdomains for DSN could break things, therefore it would be nice to have the possibility to disble it.

Note that we could make the whole sentry features list customizable in values.yaml (generated from a default feratures + "exclude_feature" and "include_feature" lists for instance). But for now maybe it's good enough as there is the need for 2 features only (`organizations:create` and `organizations:org-subdomains`)

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 6.0.0
+version: 6.1.0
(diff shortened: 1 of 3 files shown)


Mokto:
Thanks ! Could you bump the minor version please ?

PR #213, opened 2020-10-20
Add (partial) support for Symbolicator deployment

hanshsieh:
- Add partial support for Symbolicator. This should work. But the cleanup cronjob of Symbolicator is not yet implemented.
- The `/data` of Symbolicator is using *emptyDir* volume for now. (should we use PVC?) Notice the [doc](https://getsentry.github.io/symbolicator/advanced/caching/) says the volume must be POSIX-compliant.
- Increase the max size of Kafka events as minidump will exceed the default limit. I set it to the same value as [here](https://github.com/getsentry/onpremise/blob/e9cff2e28867eda078b3d3b103aae68252e19d91/docker-compose.yml#L94).
- To try it out, use
```
helm upgrade --install \
  --debug \
  --create-namespace \
  --namespace sentry \
  --wait \
  --timeout 20m \
  --set "postgresql.postgresqlPassword=changeme" \
  --set "symbolicator.enabled=true" \
  -f ./values.yaml \
  sentry ./
```
- Create a Electron project on Sentry web UI.
- And to simulate crash, clone the repo https://github.com/hanshsieh/misc/tree/crashing-electron , switch to the branch `crashing-electron`, update `main.js`, follow the `README.md`
- For the first crash report, you may need to wait a few minutes because Symbolicator will download the debug files in the background.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 6.1.0
+version: 6.2.0
 appVersion: 20.9.0
 dependencies:
   - name: redis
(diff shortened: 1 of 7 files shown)


Mokto:
Thanks! Lgtm

PR #216, opened 2020-10-25
Allow to specify strategy type recreate for sentry web deployment

### LINKED ISSUE #77: sentry-web is getting stuck in pending state after update if using PVC with ReadWriteOnce
if `filestore` is configured with PVC with access mode `ReadWriteOnce` then during update sentry-web will be stuck in pending state, as by default it will attempt to perform rolling update. It will create new instance of sentry-web, while old one is running, and new instance will never get to running state, as it will not be able to mount PVC while old instance is still running.

Fix for this is simple:
Add following to `deployment-sentry-web.yaml`:
```
  strategy:
    type: Recreate
```
That will enforce shut down of old sentry-web before creating a newer instance of sentry-web.
Possibly might be worth adding it only if `filestore.filesystem.persistence.accessMode` is set to `ReadWriteOnce`.

sagor999:
This fixes an issue for people who are using filesystem filestore with RWO type access. 
Otherwise during rolling update, existing sentry-web deployment holds access to PVC and doesn't allow new deployment to start up.
Fixes: https://github.com/sentry-kubernetes/charts/issues/77

### DIFF
--- a/sentry/templates/deployment-sentry-web.yaml
+++ b/sentry/templates/deployment-sentry-web.yaml
@@ -8,10 +8,8 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 spec:
-{{- if .Values.sentry.web.enableRecreate }}
   strategy:
-    type: Recreate
-{{- end }}
+    type: {{ .Values.sentry.web.strategyType }}
   selector:
     matchLabels:
         app: {{ template "sentry.fullname" . }}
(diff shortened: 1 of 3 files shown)


sentry/templates/deployment-sentry-web.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 spec:
+{{- if .Values.sentry.web.enableRecreate }}
+  strategy:
```
Mokto, line 11: I think we should call the value .Values.sentry.web.strategyType with a default value of RollingUpdate ?

sagor999: Done!

Mokto:
Thanks!

Mokto:
Can you please just fix the conflicts, I'll merge it afterwards

sagor999:
@Mokto done. I bumped the version to 6.2.1

PR #219, opened 2020-10-26
Update ingress-tabix.yaml

rahmanme:
got map expect array!!
to solve:

invalid type for io.k8s.api.networking.v1beta1.IngressSpec.rules: got "map", expected "array"


Mokto:
Thanks. I suppose #99 got lost along the way.

Could you bump the patch version to 7.0.1?

Thanks.

alexef:
fixed by me in: https://github.com/sentry-kubernetes/charts/pull/321/files

Mokto:
Indeed, thanks

PR #222, opened 2020-10-28
Fix missing quotes in the sentry conf

hanshsieh:
In the configmap for the Sentry config, there're some places that miss to use the `quote` function and would cause some problems in certain cases.
One example is when the admin email look like `#hello@example.com` (this is an valid email)

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 7.4.0
+version: 7.4.1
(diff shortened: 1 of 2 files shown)


hans-hsieh-asus:
Update:
- Updated the commit to also update the repo URL for stable. 
  See this announcement:  https://helm.sh/blog/new-location-stable-incubator-charts/
- Update the version in Chart.yaml

hanshsieh:
Rebased

hanshsieh:
Rebased, and remove the changes of the `Chart.yaml` that are no longer needed.

Issue #227, opened 2020-11-03
Unable to create sentry stack


Puneeth-n:
Issue: sentry-snuba-transactions-consumer, sentry-post-process-forward not coming up

Chart version: 7.1.0
Kubernetes version: 1.18
Helm version: 3

Sentry helm config
```hcl

resource "helm_release" "sentry" {
  name       = local.sentry.name
  repository = "https://sentry-kubernetes.github.io/charts"
  chart      = "sentry"
  version    = "7.1.0"
  namespace  = kubernetes_namespace.sentry.metadata[0].name
  lint       = true

  values = [
    yamlencode(
      {
        user = {
          create   = true
          email    = "lorem.ipsum@foo.com"
          password = random_password.sentry_password.result
        }

        ingress = {
          enabled = true
          annotations = {
            "kubernetes.io/ingress.class"               = "alb"
            "alb.ingress.kubernetes.io/scheme"          = "internet-facing"
            "alb.ingress.kubernetes.io/listen-ports"    = jsonencode([{ HTTP = 443 }])
            "alb.ingress.kubernetes.io/group.name"      = "infra"
            "alb.ingress.kubernetes.io/certificate-arn" = aws_acm_certificate.subdomain.arn
          }
          hostname = "sentry.${module.infra_vpc.public_subdomain}"
        }
      }
    )
  ]
}
```

kubectl -n sentry get all
```bash
NAME                                                      READY   STATUS             RESTARTS   AGE
pod/sentry-clickhouse-0                                   1/1     Running            0          50m
pod/sentry-clickhouse-1                                   1/1     Running            0          50m
pod/sentry-clickhouse-2                                   1/1     Running            0          50m
pod/sentry-clickhouse-replica-0                           1/1     Running            0          50m
pod/sentry-clickhouse-replica-1                           1/1     Running            0          50m
pod/sentry-clickhouse-replica-2                           1/1     Running            0          50m
pod/sentry-clickhouse-tabix-75698686bc-l8js6              1/1     Running            0          50m
pod/sentry-cron-768844c868-vq4dm                          1/1     Running            1          50m
pod/sentry-ingest-consumer-577bc9db4c-8z28h               1/1     Running            2          50m
pod/sentry-kafka-0                                        1/1     Running            2          50m
pod/sentry-kafka-1                                        1/1     Running            0          49m
pod/sentry-kafka-2                                        1/1     Running            1          49m
[...]

PR #229, opened 2020-11-04
New deployments should be sent as hook as well

mariusmitrofanbostontr:
@Mokto a few more deployments need to be created as hooks as they fail if the snuba-db-init and snuba-db-migrate jobs don't run first.

It makes it "unusable" for first-time people to deploy the chart. The only workaround is to set the replicas to 0 while helm is deploying it, but that's it.

Please approve, merge and tag it to 7.2.0.

Thanks!

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 7.1.0
+version: 7.2.0
 appVersion: 20.10.1
 dependencies:
   - name: redis
--- a/sentry/templates/deployment-sentry-post-process-forwarder.yaml
+++ b/sentry/templates/deployment-sentry-post-process-forwarder.yaml
@@ -7,6 +7,11 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- if .Values.asHook }}
+  annotations:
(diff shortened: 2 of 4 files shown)


Mokto:
Thanks !

james64:
Hey @Mokto @mariusmitrofanbostontr is this known to work in scenario when
* it is initial install of a chart and
* external postgres contains data from sentry v9.1.2 to be migrated to v10?

We had a migration prepared and it "failed" in a sence that
* all parts of sentry were installed and running
* postgres data (user accounts, projects settings, releases...) were migrated ok
* but issues were **not** migrated

One of differences between our test setup and migration setup was bump in version of chart to 7.2.0. We are not yet 100% sure that this is the cause but all our experimentation so far points in that direction. It works with 7.0.0 and 7.1.0 for example.

We continue to investigate on our side but I would love to hear from you guys in a meantime.

james64:
`sentry-db-init` (hook weight 6) is running before snuba consumers (hook weights set here 19,20,21). I am trying to find source for `sentry upgrade --noinput` run in `sentry-db-init` container to see whether this can affect event migration.

james64:
[This migration](https://github.com/getsentry/sentry/blob/bbf2eca0e5b3a3d6c9342f1b44e1b4a50e213cc1/src/sentry/migrations/0024_auto_20191230_2052.py#L87) is sending events to snuba service which is not started yet in version 7.2.0. If I understand #227 correctly it should be enough to tweak weights of hooks to re-order them. Do you guys agree? @mariusmitrofanbostontr @Mokto

Mokto:
Sure, go ahead if you have to change weights!

PR #240, opened 2020-11-12
Snuba hook components started before db-init

james64:
This is needed when migrating data from sentry v9 to v10. Otherwise
system installs/starts but no events are migrated. See #229 for more info.


Mokto:
Thanks!

PR #242, opened 2020-11-13
Some fixes. Do not create data PVC when not necessary. Fix double declaration in values.yaml

Antiarchitect:
Just deployed as a replacement candidate for our old Sentry 9 chart and here is what I found:
* Fix filesystem PVC creation with filestore backend not equal to "filesystem"
* Fix double replacer: declaration in values.yaml


Mokto:
Thanks!

PR #248, opened 2020-11-18
feat: org subdomain disabled per default

Mokto:


Antiarchitect:
Was very confused by this especially when single organization is turned on by default.

PR #253, opened 2020-11-25
Fix clickhouse imagePullSecrets

justincmoy:
This fixes clickhouse imagePullSecrets by consolidating on a common format (an array of strings): https://github.com/sentry-kubernetes/charts/blob/0819e79bf71d2af2dc23a9347ee242945757a4d2/clickhouse/templates/statefulset-clickhouse.yaml#L48-L51

It removes additional references to imagePullSecrets in the clickhouse statefulset templates and updates the reference in sentry's clickhouse-init hook.


justincmoy:
This is either a backward-incompatible change (semver major) or a bugfix (semver patch version), so I erred on the side of caution.

stefanocke (commented):
I currently find no way to configure it without syntactial errors, so I would appreciate to streamline it.
One question: Why did you remove it from the statefulset-clickhouse(-replica).yaml ?  And least it looks like that in the diff...

stefanocke:
Okay, I guess I can answer my own question: It was twice in statefulset-clickhouse(-replica).yaml ... once a simple string and once as array of strings. You remived the simple string one.

justincmoy:
@Mokto can you take a look at this whenever you get a chance? Thanks!

Mokto:
I'm ok with it but could you just upgrade the major version of clickhouse ?

And we'll update the sentry version in another PR.

Thanks!

justincmoy:
> I'm ok with it but could you just upgrade the major version of clickhouse ?
> 
> And we'll update the sentry version in another PR.
> 
> Thanks!

Done!

Mokto:
Thanks ;)

PR #257, opened 2020-12-03
Update service-relay.yaml

dnmgns:
Don't specify key 'annotations' again, as that causes duplicate annotation keys.

Consider the following `values.yaml` content:
```
service:
  annotations:
    foobar: "foobar" 
```

With the current code, this would generate the following yaml when running helm template (note: I'm replacing some content with "..." for readability):
```
# Source: sentry/templates/service-relay.yaml
kind: Service
...
metadata:
  name: sentry-relay
...
  annotations:
     foobar: "foobar"
  annotations:
    foobar: foobar```


dnmgns:
Closing this as it requires changes to all templates basically.

PR #258, opened 2020-12-03
Remove duplicate service annotations

dnmgns:
Don't specify key 'annotations' again, as that causes duplicate annotation keys.

Consider the following `values.yaml` content:
```
service:
  annotations:
    foobar: "foobar" 
```

With the current code, this would generate the following duplicate annotations for service-relay.yaml, service-sentry.yaml, service-snuba.yaml and service-symbolicator.yaml(note: I'm replacing some content with "..." for readability):
```
kind: Service
...
metadata:
  name: sentry-relay
...
  annotations:
    foobar: "foobar"
  annotations:
    foobar: foobar```

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 8.0.0
+version: 8.0.1
 appVersion: 20.10.1
 dependencies:
   - name: redis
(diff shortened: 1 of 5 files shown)


Mokto:
Thanks! Could you bump the patch version ?

dnmgns:
> Thanks! Could you bump the patch version ?

The chart version - right?

Mokto:
Yes, thanks!

dnmgns:
> Yes, thanks!

Thank you for a quick merge!

PR #259, opened 2020-12-03
Fix snuba when clickhouse has no replica

mthoretton:
When `clickhouse.configmap.remote_servers.replica.backup.enabled` is set to `false` snuba fails to run jobs as it still tries to connect to `sentry-clickhouse-replica-headless`.

As I understand, for non production or super critical environements `clickhouse.configmap.remote_servers.replica.backup.enabled` could be disabled without big risk, especially if clickhouse itself has a few k8s replicas.


Mokto:
Thanks!

PR #264, opened 2020-12-18
Add support of GCS without secret, using GKE ServiceAccount

tkukushkin:
We auth to GCS usgin GKE ServiceAccount with `iam.gke.io/gcp-service-account` annotation. In this case secret and environment variable for it are not needed.

https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity


Mokto (approved):

sfc-gh-pkommini:
@Mokto: Is there a AWS version of this?

Issue #166, opened 2020-08-31
Traefik 2.2 Ingress Issues


achichenin:
It's not directly an issue of this chart, yet it can be handled internally or, at least, explicitly documented.

**Routes**

Traefik 2.2.x Ingress does require path to be configured as
-  `'/api/{[1-9][0-9]*}/{(.*)}'`
 instead of 
- `'/api/[1-9][0-9]*/(.*)'`

I do not believe that there is a universal version of the configuration for all modern ingress controllers, so helm charts generally have to adapt via explicit ingress class definition in `values` and conditional template rendering with specifics.

The simplest workaround, for now, is to explicitly set route matches in the ingress configuration. 

**CSRF**

Due to issues with csrf I've had to tweak the sentry primary config-map as follows:

- add `system.url-prefix: "sentry.domain.my"` into `config.yml` section
- add `CSRF_TRUSTED_ORIGINS = ['sentry.domain.my']`  into `sentry.conf.py` section

It might be a good idea to add this to the chart as well.

Mokto:
Thanks for this ;)

Maybe you could add this to the README ?

Mokto:
Fixed with 8.1.1, Thanks @szeber

daniel5gh:
Hi, I am running into the same issue with the latest released chart.

I don't see any change/code containing `CSRF_TRUSTED_ORIGINS` either, did this get lost?

I did add `CSRF_TRUSTED_ORIGINS` to my setup manually and it resolved the issue.

thanks!

daniel5gh:
> Hi, I am running into the same issue with the latest released chart.
> 
> I don't see any change/code containing `CSRF_TRUSTED_ORIGINS` either, did this get lost?
> 
> I did add `CSRF_TRUSTED_ORIGINS` to my setup manually and it resolved the issue.
> 
> thanks!

I now changed this to use

```python
SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
SESSION_COOKIE_SECURE = True
CSRF_COOKIE_SECURE = True
SOCIAL_AUTH_REDIRECT_IS_HTTPS = True
```

instead. As this [seems the intended method](https://github.com/sentry-kubernetes/charts/blob/8.1.2/sentry/templates/configmap-sentry.yaml#L274) and forcing `CSRF_TRUSTED_ORIGINS` could actually open up a security issue.

TimShilov:
> I now changed this to use
> 
[...]

PR #274, opened 2020-12-30
Make the relay paths in the ingress compatible with traefik

szeber:
Fix for issue #166, introduces a new option under the ingress called `regexPathStyle` which defaults to `nginx` and if it's set to `traefik` it will generate the path with the regex in it in a format that traefik accepts.


Mokto:
Smart, thanks!

PR #279, opened 2021-01-14
automatically enable regex for nginx ingress controller

timfeirg:
without this piece of configuration, traffic cannot be correctly routed to sentry-relay.

this is all the issues and posts that I came across during the debug process, for reference's sake:

* https://github.com/getsentry/onpremise/issues/447
* https://github.com/getsentry/onpremise/issues/598
* https://forum.sentry.io/t/issue-with-csrf-token-after-sentry-upgrade/11133/5
* https://forum.sentry.io/t/events-get-forbidden-status-on-kubernetes/12219/7
* https://forum.sentry.io/t/event-submission-rejected-by-csrf/10482

### DIFF
--- a/sentry/templates/ingress.yaml
+++ b/sentry/templates/ingress.yaml
@@ -9,6 +9,9 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
  annotations:
+   {{- if eq (default "nginx" .Values.ingress.regexPathStyle) "nginx" }}
+     nginx.ingress.kubernetes.io/use-regex: "true"
+   {{- end }}
    {{- range $key, $value := .Values.ingress.annotations }}
      {{ $key }}: {{ $value | quote }}
    {{- end }}


timfeirg:
@Mokto

Mokto:
Sorry about the delay, and thanks!

I'm not sure it's really needed as it's explained here :
https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L323-L325

This way you're forcing this annotation for everyone.

You could use https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L321 this boolean to only set it for nginx ingress users.

timfeirg:
I overlooked and this pr is indeed not needed, sorry.

I'll try to re-write so that ingress annotations gets automatically rendered when nginx ingress is enabled. I believe this will help future users to avoid

timfeirg:
pr is updated as described in https://github.com/sentry-kubernetes/charts/pull/279#issuecomment-765155034

@Mokto

PR #280, opened 2021-01-15
do not reference CLICKHOUSE_PORT since it's set by kubernetes

timfeirg:
when deployed on kubernetes along with Clickhouse, the CLICKHOUSE_PORT
variable will by default be something like tcp://x.x.x.x:x, which cannot
be converted to str, and has no meaning in sentry helm chart config


Mokto:
Thanks!

PR #288, opened 2021-01-26
Add google sso

zifeo:
User then needs to login and enable the provider in the organization settings > auth.


Mokto:
Thanks!

mjameeghouri:
How to choose provider? @zifeo

PR #292, opened 2021-02-03
Fix default installation

NandoTheessen:
Hi everyone 👋🏻 

Due to the failing hook, an installation w/ default values is impossible at the moment. 
I appreciate this is only one of the two routes we could take to fix this issue.

I've chosen this route because I think it isn't necessary to have 6 click house pods running with the default install.

Fixes issue #283


Mokto:
Hey, thanks! I never had any issue, but I can understand.

I think we need:
1) To make this configurable ?
2) To increment the major version, as this is a breaking change.

NandoTheessen:
Hey @Mokto thanks for the super quick answer 😄

re 1: This should already be configurable in the clickhouse chart, for me this works out of the box. Is that what you meant?
re 2: Makes sense! 💡

Mokto:
My next answer is not that fast ;) Thanks for your answer, I'll merge & create a major release now.

PR #300, opened 2021-02-11
Fix missing env(CLICKHOUSE_HOST) in snuba config

phenixdotnet:
Fix missing env(CLICKHOUSE_HOST) in snuba config which prevent migration to be run correctly.

Migration job container [use the `CLICKHOUSE_HOST` env variable ](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/hooks/snuba-migrate.job.yaml#L50)to target specific host but this variable isn't used in the snuba configuration. This lead to a situation where migration are run on random host and create an issue where some host aren't up to date.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 9.0.0
+version: 9.0.1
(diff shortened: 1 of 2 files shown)


Mokto:
Thanks, could you update the patch version ? 

Thanks.

phenixdotnet:
Sure, done :)

PR #303, opened 2021-02-15
echo to stdout to understand when clickhouse-init job started/finished

azhurbilo:
Echo message to stdout to understand when clickhouse-init job started and finished

Usecase: to find via Loki (or other log aggregator) that job was triggered and finished

clickhouse-init

with empty stdout it's quite hard to understand deploy behviour


Mokto:
Thanks!

PR #305, opened 2021-02-15
Add note about scheme for system.url-prefix

daniel5gh:
Not including a scheme leads to multiple subtle issues as described in https://github.com/getsentry/sentry/issues/23365

I think it can save users a headache if they are instructed to add the scheme to the URI config value. It's called URI and that might imply it needs a scheme, but I just had a hostname and that lead to multiple issues where sentry was using _some_ URIs that has a double hostname.

ref: https://develop.sentry.dev/config/#general

### DIFF
--- a/sentry/values.yaml
+++ b/sentry/values.yaml
@@ -230,6 +230,7 @@ hooks:
         memory: 1Gi
 
 system:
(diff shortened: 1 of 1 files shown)


Mokto:
Thanks!

PR #306, opened 2021-02-16
feat: update sentry to february release

Mokto:


azhurbilo:
1) Do we need to change symbolicator version to 3.3.0 like described here https://github.com/getsentry/onpremise/compare/21.1.0...21.2.0 ?


2) And remove slack legacy field

> Unknown config option found: 'slack.legacy-app'

according to https://github.com/getsentry/sentry/pull/23587 ?

Mokto:
Good point! Thanks.

But for Slack legacy, it still seems to be working for me.

PR #307, opened 2021-02-16
set failureThreshold for readiness and liveness probe

azhurbilo:
liveness failureThreshold should be more than readiness

otherwise container will be restarted before deciding that container not ready


Mokto:
Thanks! Maybe you could also make those values changeable in `values.yaml` ?

PR #309, opened 2021-02-18
Fix BROKER_URL for rabbitmq auth

justincmoy:


Mokto:
Thanks! Could you also update the patch version of the chart ?

justincmoy:
> Thanks! Could you also update the patch version of the chart ?

Done!

Mokto:
Thanks!

PR #317, opened 2021-02-26
feat: allow setting tabix as the default url, 

alexef:
using https://clickhouse.tech/docs/en/operations/server-configuration-parameters/settings/#server_configuration_parameters-http_server_default_response

note: I spent quite some time trying to overcome the CORS issues with tabix and CH using different domain names, and I eventually give up. this works


alexef:
I believe I found a different way to do it, by putting tabix under a sub-path, on an ingress with the same host as ch

PR #321, opened 2021-03-01
feat: allow setting ingress annotations for dealing with cors

alexef:
- allow setting ingress annotations for dealing with cors
- fix ingress service port name (http-port)
- fix ingress syntax - tabix

I needed these changes in order to be able to connect from tabix to clickhouse using different domain names (cors)


clickhouse/templates/ingress-tabix.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 spec:
   rules:
-    host: {{ .Values.tabix.ingress.host }}
+  - host: {{ .Values.tabix.ingress.host }}
```
Mokto, line 13: I think you need to apply the same thing to hosts in tls ? 

Cf https://github.com/sentry-kubernetes/charts/pull/219/files

alexef: you are correct, I didn't notice that.

alexef: done

Issue #270, opened 2020-12-23
Deploy fails with `asHook=false`


paulshannon:
I believe this is a duplicate of the unresolved but closed #217 

This is using a clean local copy of the December release

```
$ helm upgrade --install --namespace sentry -f sentry.values.yml --set asHook=true sentry ./charts/sentry
Release "sentry" does not exist. Installing it now.
NAME: sentry
LAST DEPLOYED: Fri Dec 18 08:29:16 2020
NAMESPACE: sentry
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
* When running upgrades, make sure to give back the `system.secretKey` value.

kubectl -n sentry get configmap sentry-sentry -o json | grep -m1 -Po '(?<=system.secret-key: )[^\\]*'
$ helm upgrade --install --namespace sentry -f sentry.values.yml --set asHook=false sentry ./charts/sentry
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: Deployment "sentry-relay" in namespace "sentry" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "sentry"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "sentry"
```

Running an otherwise unmodified upgrade with asHook=False immediately after deploy (as suggested [here](https://github.com/sentry-kubernetes/charts/issues/217#issuecomment-718821943)) also fails

Mokto:
@paulshannon could you check again ? @jandersen-plaid has pushed a fix.

jandersen-plaid:
To contextualize the fix a bit: you will need to freshly install the chart or manually add the labels and annotations to the deployments in #217 for the fix to work -- the error will still happen if you are trying to use an older version.

This should work:
```
$ helm upgrade --install --namespace sentry -f sentry.values.yml --set asHook=true sentry ./charts/sentry --version 9.1.2
$ helm upgrade --install --namespace sentry -f sentry.values.yml --set asHook=false sentry ./charts/sentry --version 9.1.2
```

Or this:
```
$ export SENTRY_NAMESPACE=sentry
$ helm upgrade --install --namespace ${SENTRY_NAMESPACE} -f sentry.values.yml --set asHook=true sentry ./charts/sentry --version 9.1.1
[...]

Issue #217, opened 2020-10-26
Correct labels/annotations for "asHook" deployments


james64:
Hello,

initial installation (`asHook: true`) is going through ok. Upgrades however (`asHook: false`) are not coming through because of Helm error (helm v3.3.4):
```
"Error: UPGRADE FAILED: rendered manifests contain a resource that already exists.
Unable to continue with update: Deployment "sentry-relay" in namespace "infra-sentry" exists and cannot
be imported into the current release: invalid ownership metadata; label validation error: missing key
"app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key
"meta.helm.sh/release-name": must be set to "sentry"; annotation validation error: missing key
"meta.helm.sh/release-namespace": must be set to "infra-sentry"
```

Only way to proceed is to manually add label and annotation in cluster. This is happening for all resources affected by `asHook` parameter:
- sentry-relay
- sentry-snub-outcomes-consumer
- sentry-snuba-replacer
- sentry-sessions-consumer

Other resources are correctly labeled/annotated. Can we add these to hook deployments as well? Or is there any reason why not to?

Thank you for this chart!

Mokto:
Yes, let's add them if necessary. Never got any issue though ;)

PR welcomed!

james64:
I am happy to provide PR for this. I am going to hold it for now. I asked about this on helm side. Let's see what they have to say here.

james64:
Helm guys says what I see is expected. What is still interesting for me is that sometimes I do see the error and sometimes it is ok and I do not need to do any manual edits. I will observe this for a while so I can understand better what is going on.

james64:
When one wants to do some configuration change and switch `asHook` from true to false at the same time that is when described helm issue occurs. Only way how to not have issues with this is to do upgrade with `asHook=false` immediately after initial deploy **without** any changes. On the second run helm adds these annotations automatically. So we do not need to add these labels manually. Maybe include this info to comment above `asHook` value only. Thanks

james64:
Even without any changes to user configuration this has the effect of removing helm hook annotations which is enough to cause helm crash from the beginning. We do not necessarily need to re-open this. Just if you guys have some way around this I would love to hear about it. And I guess you do since you "never got any issue" like this :)

paulshannon:
I am also running into this problem but it appears to be every time I attempt to upgrade.

```
[...]

PR #322, opened 2021-03-03
Add labels and annotations to let upgrades from asHook True to False to succeed

jandersen-plaid:
This fixes issues:

* https://github.com/sentry-kubernetes/charts/issues/217
* https://github.com/sentry-kubernetes/charts/issues/270

When installing with `asHook=true` we add the mandatory Helm annotations to the necessary deployments, so that upgrading the value to `asHook=false` does not fail.

Additionally, we set the label `app.kubernetes.io/managed-by: "Helm"` as Helm should always manage these deployments.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 9.1.1
+version: 9.1.2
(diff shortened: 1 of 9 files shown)


Mokto:
Cool, thanks. Could you upgrade to 9.1.2 please ?
I just merged another PR.

jandersen-plaid:
For sure!

PR #323, opened 2021-03-03
Reference the kebab case deployment consumer that exists

jandersen-plaid:
The HPA for the ingest-consumer references the wrong deployment (which causes autoscaling for the ingest-consumer to be ineffective). The deployment used in the chart is `-ingest-consumer` and not `-ingestConsumer`.

This PR fixes the reference to the deployment name in the HPA yaml.


Mokto:
Thanks!

PR #324, opened 2021-03-04
Add support for sidecars and volumes

alexef:
This is a bit involved, but let me explain the situation:

- we're using AWS Elasticcache for Redis with in-transit and at-rest encryption enabled
- while some parts of sentry allow using `rediss` as the protocol, others don't
- the only option for us is injecting an `stunnel` (not part of this PR) container that runs next to the sentry code and proxies Redis requests to Elasticcache

Thank you for considering the change.

### DIFF
--- a/sentry/README.md
+++ b/sentry/README.md
@@ -65,8 +65,6 @@ Parameter                          | Description                                
 `sentry.features.vstsLimitedScopes` | Disables the azdo-integrations with limited scopes that is the cause of so much pain | `true`
 `sentry.web.customCA.secretName` | Allows mounting a custom CA secret | `nil`
 `sentry.web.customCA.item` | Key of CA cert object within the secret | `ca.crt`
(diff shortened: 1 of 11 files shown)


Mokto:
Hey! Thanks. I'm not against it, but I would separate sidecars&volumes per application. 
Right now you have to set the same sidecars for all deployments.

alexef:
@Mokto thanks for reviewing this!

I'm a newbie in helm, I'll refactor this to be per application instead. What I would like to avoid in our own `values` is that we have to repeat the same sidecar/volume configuration 7 times. Will look if there is an easy way to reuse a value. Hints on how I can define it like that are welcome :)

alexef:
@Mokto updated. Couldn't find a way to re-use values, so I'll just duplicate them. Thank you for having another look.

Mokto:
Thanks!

Mokto (approved):

PR #328, opened 2021-03-13
Added support for aws alb ingress controller

phenixdotnet:
Hi.
Please find a proposition to support [AWS Load balancer controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller) directly from the chart configuration.

The AWS LB support is enabled by setting `regexPathStyle` to `aws-alb`

### DIFF
--- a/sentry/templates/ingress.yaml
+++ b/sentry/templates/ingress.yaml
@@ -53,7 +53,6 @@ spec:
               serviceName: {{ template "sentry.fullname" . }}-web
               servicePort: {{ .Values.service.externalPort }}
     {{- end }}
-    {{- end }}
 {{- if .Values.ingress.tls }}
   tls:
 {{ toYaml .Values.ingress.tls | indent 4 }}
--- a/sentry/templates/ingress.yaml
+++ b/sentry/templates/ingress.yaml
@@ -22,8 +22,7 @@ spec:
             backend:
               serviceName: {{ template "sentry.fullname" . }}-nginx
               servicePort: {{ .Values.nginx.service.port }}
-    {{- else }}
-    {{- if eq (default "nginx" .Values.ingress.regexPathStyle) "aws-alb" }}
+    {{- else if eq (default "nginx" .Values.ingress.regexPathStyle) "aws-alb" }}
           - path: /api/0/*
             backend:
               serviceName: {{ template "sentry.fullname" . }}-web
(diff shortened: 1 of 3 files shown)


sentry/templates/ingress.yaml
```diff
@@ -22,6 +22,20 @@ spec:
             backend:
               serviceName: {{ template "sentry.fullname" . }}-nginx
               servicePort: {{ .Values.nginx.service.port }}
+    {{- else }}
```
Mokto: I think using `else if` would be easier to read: https://helm.sh/docs/chart_template_guide/control_structures/#ifelse

Mokto (changes requested):

sentry/templates/ingress.yaml
phenixdotnet: okay no prob :)

sentry/templates/ingress.yaml
```diff
@@ -22,6 +22,19 @@ spec:
             backend:
               serviceName: {{ template "sentry.fullname" . }}-nginx
               servicePort: {{ .Values.nginx.service.port }}
+    {{- else if eq (default "nginx" .Values.ingress.regexPathStyle) "aws-alb" }}
```
joel-schaal, line 25: Wouldn't that be suitable for all cases (alb, traeffic,..) ?
I think the paths defined in that order remove the need for any RegEx.

phenixdotnet: Not sure about how other ingress controller process the paths.
AWS alb is very ... limited, it process it in the configuration order but if nginx or traefik process it from more restrictive to less restrictive, this *could* not work as expected. 
And the main problem is still here: Each ingress controller has it's own regex syntax so anyway we have to deal with it.

sentry/templates/ingress.yaml
```diff
@@ -40,6 +53,7 @@ spec:
               serviceName: {{ template "sentry.fullname" . }}-web
               servicePort: {{ .Values.service.externalPort }}
     {{- end }}
+    {{- end }}
```
Mokto: This is extra right ?

Mokto (changes requested):

sentry/templates/ingress.yaml
phenixdotnet: Yes you're right ^^'

Mokto (approved):

phenixdotnet:
Thanks !

PR #331, opened 2021-03-15
Snuba requires clickhouse 20.3.9.70

ebuildy:
reference https://github.com/getsentry/onpremise/issues/750


Mokto:
Hi. Thanks! But the right snuba version is already set here : https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L402

PR #334, opened 2021-03-16
Allow to use authorization in clickhouse

turbotankist:
Clickhouse need to be protected by auth,
So there is a PR with it.
Clickhouse has default user "default" with empty ("") password


sentry/templates/_helper.tpl
```diff
@@ -258,6 +258,53 @@ Set ClickHouse port
 {{- end -}}
 {{- end -}}
 
+{{/*
+Set ClickHouse Databse
```
J0sh0nat0r: nit: database

sentry/values.yaml
```diff
@@ -406,6 +406,18 @@ clickhouse:
...
             enabled: false
+      users:
+        enabled: false
+        user:
+        # the first user will be used if enabled
+        - name: default
+          config:
+            password: ""
+            networks:
+            - ::/0
```
J0sh0nat0r, line 417: Should we include `0.0.0.0/0` as a default here? The IPv6 rule didn't seem sufficient to allow access when testing locally. (not that users shouldn't put their cluster subnets here anyway, maybe worthy of a comment?)

J0sh0nat0r (approved):
LGTM 👍 Thanks for the PR!
Terribly sorry it took me this long to get around to reviewing this.

sentry/values.yaml
turbotankist: I just copied default values from 
https://github.com/sentry-kubernetes/charts/blob/develop/clickhouse/values.yaml#L324
Also ::/0 usually mean both IPv4 and IPv6 subnetworks
https://clickhouse.tech/docs/en/operations/settings/settings-users/  
```
To open access for user from any network, specify:
<ip>::/0</ip>
```

sentry/templates/_helper.tpl
turbotankist: fixed

Mokto:
Thanks @J0sh0nat0r @turbotankist

PR #335, opened 2021-03-16
Pass symbolicator config as YAML

philicious:
This adds more config options to symbolicator deployment.

We needed `connect_to_reserved_ips` , which wasnt supported, so I took the chance and added more along the way.

I tested symbolicator to start properly also with the `caches` enabled

### DIFF
--- a/sentry/README.md
+++ b/sentry/README.md
@@ -31,6 +31,8 @@ If Redis auth is disabled:
 
 The following table lists the configurable parameters of the Sentry chart and their default values.
 
+Note: this table is incomplete, so have a look at the values.yaml in case you miss something
+
 Parameter                          | Description                                                                                                | Default
 :--------------------------------- | :--------------------------------------------------------------------------------------------------------- | :---------------------------------------------------
 `user.create` | if `true`, creates a default admin user defined from `email` and `password` | `true`
@@ -65,6 +67,8 @@ Parameter                          | Description                                
 `sentry.features.vstsLimitedScopes` | Disables the azdo-integrations with limited scopes that is the cause of so much pain | `true`
 `sentry.web.customCA.secretName` | Allows mounting a custom CA secret | `nil`
 `sentry.web.customCA.item` | Key of CA cert object within the secret | `ca.crt`
+`symbolicator.api.enabled` | Enable Symbolicator | `false`
+`symbolicator.api.config` | Config file for Symbolicator, see [its docs](https://getsentry.github.io/symbolicator/#configuration) | see values.yaml
 
 ## NGINX and/or Ingress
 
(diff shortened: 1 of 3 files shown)


sentry/templates/configmap-symbolicator.yaml
```diff
@@ -14,8 +14,21 @@ data:
     cache_dir: "/data"
     bind: "0.0.0.0:{{ template "symbolicator.port" }}"
     logging:
-      level: "{{ .Values.symbolicator.api.logging.level }}"
+      level: {{ .Values.symbolicator.api.logging.level | quote }}
     metrics:
-      statsd: null
-    sentry_dsn: null # TODO: Automatically fill this with the internal project DSN
+    {{- if .Values.symbolicator.api.statsd.server }}
```
Mokto: Maybe we should make this logic super simple by just passing a yaml from the values file ?
With toYaml

It feels we need to add each config in the repository every time a new one is available.

philicious: Actually true. Will update the PR in next days

philicious: @Mokto I updated the PR. Symbolicator starts fine with the default values config and also works for me when passing custom config.
should be a backwards compatible change as the defaults didnt change

philicious:
@Mokto I could add some smybolicator paragraph to README as it wasnt totally straight-forward to get it working, like
```
        filestore:
          filesystem:
            persistence:
              persistentWorkers: true
```
but also needed to set `connect_to_reserved_ips: true` 

maybe its even worth changing `connect_to_reserved_ips` to default true?

Mokto:
I agree with both.

I'll let you update your PR and then I'll merge it.


Thanks!

philicious:
updated README and also default value of `connect_to_reserved_ips` @Mokto

sentry/README.md
```diff
@@ -77,3 +81,23 @@ For your security, the [`system.secret-key`](https://develop.sentry.dev/config/#
...
+
+## Symbolicator
+
+For getting native stacktraces and minidumps symbolicated with debug symbols (e.g. iOS/Android), you need to enable Symbolicator via
+```yaml
+symbolicator:
+  enabled: true
+```
+
+However, you also need to share the data between sentry-worker and sentry-web via
```
Mokto: This is not the recommended way. 
Recommended way is to use something else than filestore, like S3 or GCS.

Please adapt your description with this ;)

philicious: hm yes I agree. however S3 is not supported by the chart. Another option I forgot to mention is using an AWS EFS Storageclass, which supports ReadWriteMany.
I will update the doc and make the options more clear

Mokto: https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L359
https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L398-L405

S3 is supported by the chart 👍

philicious: Hm ye I saw it in values but imho I didn't find a reference in the templates or did I miss it?
(Wouldn't have worked in my personal case tho as we cannot use IAM users/access-keys but need assumable roles and so on)

philicious: @Mokto I updated again to be more precise about the storage options and caveats

Mokto:
Thaaaanks ! 👍

PR #348, opened 2021-04-08
Fix podAnnotations

klengyel:
This fixes the default values for pod annotations (silly type mismatch) and also includes it for clickhouse-init job.


Mokto:
Thanks!

PR #353, opened 2021-04-14
Change default relay mode to managed.

Antiarchitect:
Use relay managed mode by default in order to fix:
https://github.com/getsentry/sentry/issues/22789
https://github.com/getsentry/sentry/issues/23161


ckoehn:
I was about to make the same PR. Sentry does basically no data scrubbing with the current defaults!

Mokto:
Let me try it today but I had a lot of issues previously with this mode.

Antiarchitect:
It's about a week we turned this mode on for our corporate Sentry - all is OK for now. Rate Limits and DSN checking now works at last.

Mokto:
Thanks!

PR #356, opened 2021-04-18
feat: 21.4.0 release

alekitto:
April 2021 release


Mokto:
Thanks for your work on that.

Do we really need to upgrade the major version? I don't think anything is breaking.

I'll try it today.

alekitto:
@Mokto ok, I modified the `version` field to increment only the minor version

Mokto:
Thanks! Everything seems to work fine on my end.

PR #357, opened 2021-04-20
Fix cleanup job

alekitto:
Related to #356
 
I misinterpreted a commit from `onpremise` repo, so there was an error on the cleanup cronjob. This should revert the erroneous change.
Also, I noted that snuba cleanup jobs are not present. Should I add them?


Mokto:
Sure, would be nice! ;)

alekitto:
Ok @Mokto, added the snuba cleanup cronjobs. I copied the interval from the `onpremise` repo, but IMHO it is a little too frequent. What do you think about it?

Mokto:
I don't know how that works internally but yeah that seems extreme. Maybe every hour then ?

alekitto:
> Maybe every hour then ?

Seems reasonable, I've updated the default values accordingly

zifeo:
This will fix #362

PR #359, opened 2021-04-21
Fix dbCheck image can be pulled from custom repository (#358)

### LINKED ISSUE #358: Can't deploy sentry to air gapped environment - sentry-db-check job is using subfuzion/netcat image from dockerhub
Hello,

My problem:
I can't deploy Sentry in air gapped environment, since sentry-db-check job (helm hook) is using `subfuzion/netcat:latest` image directly from docker hub.

Proposed solution:
Add custom definition to add this image as a value, so this chart can be fully used in air gapped env.  

PS: I know that I can disable hooks, but that would disable also the dbInit and snubaInit jobs, so I don't see that as an option.
PSS: I can send a PR

Thank you

vojtechmares:
This allows Sentry to be installed with this helm chart in air gapped
environment

Closes #358

### DIFF
--- a/sentry/templates/_helper.tpl
+++ b/sentry/templates/_helper.tpl
@@ -35,6 +35,12 @@
 {{- .Values.images.symbolicator.tag -}}
 {{- end -}}
 
+{{- define "dbCheck.image" -}}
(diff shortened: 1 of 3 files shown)


Mokto (approved):

Mokto:
thanks!

PR #368, opened 2021-04-29
Fix netcat variable image evaluation

mleclerc00:
Adding the missing . for allowing top-level scope when evaluating the dbCheck.image variable

Without this the image override defined in values.yaml doesn't work properly

https://github.com/sentry-kubernetes/charts/blob/05698720bc8a5eedcbaa847f766ac6a669360ec5/sentry/values.yaml#L241


Mokto:
Thanks

PR #387, opened 2021-05-18
Add http port parameter for external clickhouse

alekitto:
Without setting `CLUSTERS[0]["http_port"]` the default 8123 is used, leading to a `Connection refused` error when a non-default port is used by clickhouse server.


Mokto:
Thanks.

PR #389, opened 2021-05-21
upgrade sentry to 21.5.1

alekitto:
While upgrading to 21.5 I encoutered some problems with `DEFAULT_BROKERS` setting which was not loaded from the `settings.py` in config map.

Inspecting the source code of snuba I discovered that this setting is now loaded from env vars, so I moved it into the env block and de-duplicated all the snuba env var blocks into an helper.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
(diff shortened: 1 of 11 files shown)


Mokto:
Thanks! I was going to create the same in 30min 👍 

Could you upgrade the minor version instead of the patch version?

Thanks!

alekitto:
Updated to 11.3.0

PR #394, opened 2021-05-26
Add affinity, nodeSelector and tolerations to values of hooks

sue445:
Deprecated chart supported `affinity` and `tolerations` in hooks. 

https://github.com/helm/charts/blob/f4f301ae450101b981805bd045451f08c0d74afa/stable/sentry/templates/hooks/db-init.job.yaml#L31-L38

So I added a commonly used selector.


Mokto:
Thanks!

PR #395, opened 2021-05-27
Fix snuba metrics collection

andrsp:
Add statsd collector host/port to snuba config if metrics enabled in chart

It defaults to localhost:8125 and produces lots of error messages in shuba components logs (.. dropping packet) if statsd is not available, so lets configure it right if it’s available :)


Mokto:
Thanks!

PR #398, opened 2021-05-28
ALB Ingress : Fix annotations parsing error

LouisPolyconseil:
Fix parsing error when you enable alb http redirect and have custom annotations at the same time


Mokto:
Thanks.

PR #404, opened 2021-06-08
Add max_server_memory_usage to clickhouse config

sue445:
I want to use [`max_server_memory_usage`](https://clickhouse.tech/docs/en/operations/server-configuration-parameters/settings/#max_server_memory_usage), but it wasn't in the helm setting.

So I added.


clickhouse/templates/configmap-config.yaml
```diff
@@ -40,6 +40,7 @@ data:
         <max_session_timeout>{{ .Values.clickhouse.configmap.max_session_timeout | default "3600" }}</max_session_timeout>
         <default_session_timeout>{{ .Values.clickhouse.configmap.default_session_timeout | default "60" }}</default_session_timeout>
         <disable_internal_dns_cache>{{ .Values.clickhouse.configmap.disable_internal_dns_cache | default "1" }}</disable_internal_dns_cache>
+        <max_server_memory_usage>{{ .Values.clickhouse.configmap.max_server_memory_usage | default "0" }}</max_server_memory_usage>
```
Mokto, line 45: Should we make it completely optional by not setting this value at all ?

[The default values seems to be 10GB](https://clickhouse.tech/docs/en/operations/settings/query-complexity/#settings_max_memory_usage) ?

sue445: @Mokto I fixed at 04aa903

Mokto:
Thanks!

PR #380, opened 2021-05-12
configurable service account and existingSecret for externalPostgresql

kimxogus:
For EKS IRSA, this is necessary

PR #405, opened 2021-06-08
Add possibility to configure service accounts

Phylu:
This enables this helm chart to be used in an environment where service accounts are needed such as a Kubernetes cluster using Consul with ACLs enabled.

Currently, when trying to use the helm chart when Consul ACLs are enabled there is no possibility to get it running as the pods cannot be created.  For more information on why this is needed in this case, see: https://github.com/hashicorp/consul-helm/issues/202).

The nginx chart update is needed as the possibility to configure Service Accounts is not available in the old version. According to its documentation, a dependency update is important (https://github.com/bitnami/charts/tree/master/bitnami/nginx#to-700). I am not sure how exactly the lifecycle and versioning of the dependencies is handled here. Any pointers are appreciated in case I need to adjust something for that.

As I am still testing the setup on a Consul enabled Kubernetes cluster, there may be some smaller adjustments still incoming. However, I really hope to get some feedback already as there were quite some changes needed to get it running.


Phylu:
I did some more testing on my setup and everything seems to be working as expected now :)

z0rc:
Supersedes #380 (which doesn't provide template for ServiceAccount creation and uses single SA for all deployments/pods). 

This PR would also allow to use S3 for filestorage without hardcoding AWS credentials, auth would be done via ServiceAccount and IAM role associated by SA annotations.

@Mokto any update on status of this PR?

Phylu:
> Supersedes #380 (which doesn't provide template for ServiceAccount creation and uses single SA for all deployments/pods).
> 
> This PR would also allow to use S3 for filestorage without hardcoding AWS credentials, auth would be done via ServiceAccount and IAM role associated by SA annotations.
> 
> @Mokto any update on status of this PR?

Just merged in Upstream to resolve merge conflicts. So from my side, this is good to go.

Mokto (approved):

Mokto:
Thanks!

sentry/values.yaml
```diff
@@ -630,6 +649,15 @@ kafka:
...
     port: 9092
 
+  serviceAccount:
+    create: false
+    name: kafka
+
+  zookeeper:
+    serviceAccount:
+      create: false
+      name: zookeeper
```
philstutz, line 659: This changes the current behavior. The zookeeper chart will use the `name` even with `create: false`, so to use the current `default`, it needs to be overridden. Also for redis below, the other charts don't use the `name` if `create` is `false`.

Phylu: Sorry for that.
I created a new PR which adds comments around these configurations by default so it should reset the default behavour from before: https://github.com/sentry-kubernetes/charts/pull/447

philstutz: Thanks!

PR #409, opened 2021-06-09
fix: Make the snuba api liveness and readiness timeouts configurable

### LINKED ISSUE #408: Snuba API timeout too short -- CrashLooping under load
We have been scaling up our k8s Sentry recently and, as it has taken on more users and load, we are finding that the snuba API pods have a habit of going into `CrashLoopBackOff`. Looking further into this issue, we have found that the issue only comes up when the snuba-api pods are undergoing high load and always happens because the pods miss liveness and readiness checks. Specifically, while under high load, the snuba-api pods will timeout when responding to the kube-probe, but will respond to `/query` requests in a reasonable amount of time.

The pods themselves are behaving fine -- serving query requests normally -- and the resource usage for the pods look fine:
* CPU is under 20~ish% and not being throttled
* Memory is around 30%

but it looks like there might be some blocking behavior in the WSGI responding to a high volume of requests to `/query`.

Now, this isn't something that is super fixable, and only happens when the snuba-api pods are under just enough load to hit the HPA scaling thresholds (and even out that high load) but have not scaled out yet. 

For that reason, I think the snuba api readiness and liveness timeouts should be configurable -- it is a pretty easy fix, so I will likely open a PR against it, but I am definitely open for other solutions here!

jandersen-plaid:
Fixes #408 by making the liveness and readiness timeouts for snuba-api configurable.

Tested it out via templating and it went smoothly, but let me know if this option add is unwarranted!

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 11.3.2
+version: 11.3.3
 appVersion: 21.5.1
 dependencies:
   - name: redis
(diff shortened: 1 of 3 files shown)


Mokto:
Thanks!

Issue #410, opened 2021-06-10
Redis password containing certain special characters breaks deployment


johntwyman:
I tried deploying the sentry chart to our cluster making use of the `externalRedis` values. We have an existing Redis service on our cluster so it makes sense to take advantage of it.

Unfortunately the password for our Redis service includes a `/` character. This breaks the sentry deployment.

The issue is that the redis password is not passed through the Helm function `urlquery` when it's being used in the redis URLs in the snuba and sentry configmap template files.

I've tested passing the value through `urlquery` in the `charts/_helper.tpl` file but that led to other errors, as `sentry.redis.password` gets used in a couple of ways beyond construction of `redis://:<pass>@<host>:<port>` style URLs.

But I have modified the construction of the redis URLs in the sentry and relay config map template files with success. The deployment has worked. I'm happy to raise a PR for review.

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #411, opened 2021-06-10
#410 - Sanitise redis password with urlquery function

johntwyman:
This PR addresses issue #410 - sanitising the supplied redis password with the Helm `urlquery` function to accommodate characters such as `/`. It's been tested locally and on our staging cluster with success. We can connect to our external redis service using a complex password and we've seen no evidence of other errors in the logs of the various service components.

### DIFF
--- a/sentry/templates/configmap-relay.yaml
+++ b/sentry/templates/configmap-relay.yaml
@@ -30,7 +30,7 @@ data:
           value: 50000000  # 50MB or bust
 
       {{- if $redisPass }}
-      redis: "redis://:{{ $redisPass }}@{{ $redisHost }}:{{ $redisPort }}"
+      redis: "redis://:{{ $redisPass | urlquery }}@{{ $redisHost }}:{{ $redisPort }}"
       {{- else }}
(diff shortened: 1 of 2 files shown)


Mokto:
Thanks. I'm a bit concerned about backward compatibility, though.

Maybe we should bump the major version?

johntwyman:
This PR does not introduce a breaking change. For any password that doesn't involve characters like `/` the output of `urlquery` is unchanged.

Mokto:
I'm actually talking about the people that are using special character. Maybe they've escaped the password manually in the values.yaml file.

johntwyman:
Ah gotcha! Thanks @Mokto I follow you now.

Yes I agree that's a good point. Further encoding already encoded passwords would break things.

devsaider:
> I'm actually talking about the people that are using special character. Maybe they've escaped the password manually in the values.yaml file.

I don't think its possible to use url encoded password, it would break snuba here: https://github.com/sentry-kubernetes/charts/blob/46a1256e7638c0a4d5c171889a6df695d45d9fed/sentry/templates/configmap-snuba.yaml#L59

PR #415, opened 2021-06-15
Remove extra quotation mark in ConfigMap

GoIT-FCC:
Due to this quotation mark, GitHub integration does not work, Sentry fails with "Could not deserialize key data" error

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 11.4.0
+version: 11.4.1
 appVersion: 21.6.1
 dependencies:
   - name: redis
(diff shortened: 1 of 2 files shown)


kilhage (approved):

GoIT-FCC:
@Mokto Please, review this change as it blocks our work

Mokto:
Sorry about the delay.

I had to add this hack, without it doesn't work.

This works for me :

```

  github:
    appId: "1234"
    appName: MyApp
    clientId: "yyy"
    clientSecret: "XXX"
    privateKey: "-----BEGIN RSA PRIVATE KEY-----\nAAAAAAA\BBBBB==\n-----END RSA PRIVATE KEY-----\n"
    webhookSecret: "XXXX"
```

GoIT-FCC:
@Mokto could you try with my latest changes? It works for me, but it doesn't work with your hack with extra quotation

Mokto:
Trying now. Thanks!

PR #416, opened 2021-06-16
upgrade sentry to 21.6.1

alekitto:
I've upgraded the appVersion only, as it seems that no other change is needed.


Mokto:
Thanks!

max-wittig:
Isn't there more needed? What about the `wal2json` plugin?

https://github.com/getsentry/onpremise/compare/21.5.1...21.6.0

PR #412, opened 2021-06-11
Add maxBatchSize to snuba.replacer and snuba.outcomesConsumer

sue445:
I run Sentry in my company. 

Kafka consumer is delayed because our Sentry is receiving a lot of errors...

image

As the Consumer continues to delay, we will also get an error like https://github.com/getsentry/onpremise/issues/478 on `sentry-snuba-outcomes-consumer`.

In my research, it was improved by increasing the `--max-batch-size` of` sentry-snuba-outcomes-consumer`.

image

So I want to be able to pass any value to  `--max-batch-size`

PR #418, opened 2021-06-18
Add maxBatchSize to some deployments

sue445:
This patch is similar to #412, but I added `maxBatchSize` to other than`snuba.replacer` and `snuba.outcomesConsumer` too.


Mokto (approved):

Mokto:
Thanks!

PR #420, opened 2021-06-22
Fix #419 using correct block indicator

### LINKED ISSUE #419: bash script in clickhouse-init hook gets mangled due to `>-`
the clickhouse-init script does not run properly because the use of '>-' in the definition of it messes with line-breaks in seemingly random places, changing this to `|-` fixes it.

file: charts/sentry/templates/hooks/clickhouse-init.job.yaml

dhxgit:
using the literal `|` block indicator instead of the chomping one, to preserve the bash script correctly.


Mokto:
Thanks.

PR #423, opened 2021-06-24
Fixed. `Error: no such option: -c` when `sentry.ingestConsumer.concurrency` is passed

sue445:
When I set any value in `sentry.ingestConsumer.concurrency`, the following error will occur.

```
E 2021-06-24T04:09:37.862710536Z Error: no such option: -c 
E 2021-06-24T04:09:49.563666344Z Usage: sentry run ingest-consumer [OPTIONS] 
E 2021-06-24T04:09:49.563731229Z Try 'sentry run ingest-consumer --help' for help.
```

Because `sentry run ingest-consumer` supports only `--concurrency` arg. 

https://github.com/getsentry/sentry/blob/21.6.1/src/sentry/runner/commands/run.py#L468-L474

So I fixed.


Mokto:
Thanks!

PR #432, opened 2021-07-02
Add probe arguments to sentry-web and sentry-relay

sue445:
Now, only `initialDelaySeconds` can be set in `sentry-web` and `sentry-relay` Probes.

* https://github.com/sentry-kubernetes/charts/blob/11.4.0/sentry/templates/deployment-sentry-web.yaml#L101-L120
* https://github.com/sentry-kubernetes/charts/blob/11.4.0/sentry/templates/deployment-relay.yaml#L101-L120

So I added other parameters because I want to be able to set other parameters freely.


Mokto:
Thanks ;)

PR #441, opened 2021-07-23
Add BackendConfig when GKE Ingress controller is enabled

xpicio:
Hello, i'm doing some test on this chart to deploy it on GKE (1.19.10-gke.1000). When the **Ingress** and the health checks to **sentry-web** and **sentry-relay** has been created the health check to **sentry-relay** service is misconfigured. The health check try to connect to **sentry-relay** on port 80 with path / (the default) ignoring the configuration on **readinessProbe**. There are many cases that may cause this behaviour, so the best choice on GKE is to use **BackendConfig CRD** to configure the health checks.

Here some reference from google documentations:
https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#health_checks
https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#direct_health

If you find any issue on my changes let's me know, hope this help

### DIFF
--- a/sentry/templates/_helper.tpl
+++ b/sentry/templates/_helper.tpl
@@ -9,7 +9,9 @@
 
 {{- define "nginx.port" -}}{{ default "8080" .Values.nginx.containerPort }}{{- end -}}
 {{- define "relay.port" -}}3000{{- end -}}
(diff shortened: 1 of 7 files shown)


Mokto:
Thanks!

PR #447, opened 2021-07-30
Update values.yaml

Phylu:
Ensure that default behaviour will take place when no external service accounts are wanted


Mokto:
thx

Issue #407, opened 2021-06-09
Update Ingress to use networking.k8s.io/v1 api version


z0rc:
```
networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

Helm recently updated default ingress template to accommodate new api version, see https://github.com/helm/helm/pull/9621

razuos:
This would be nice, just had problems juggling with `pathType`s and had to enable nginx for event ingestion to work.

PR #448, opened 2021-08-02
Update Ingress to use networking.k8s.io/v1 API Version

vaivanov:
Improve ingress template to add support for new ingress features added in Kubernetes 1.18 and 1.19:

`networking.k8s.io/v1` support
`ingressClassName` field support
`pathType` field support

Fix for #407


zifeo:
@Mokto Can we merge this?

Mokto:
Yes, Thanks!

PR #449, opened 2021-08-02
Fix metrics serviceAccount

z0rc:
Followup to #405

Metrics deployment used common prefix only for SA, while it should be like `{{ prefix }}-metrics`.

CC @Phylu


z0rc:
@Mokto any updates on your side?

I want to stress the importance of this change. Without the fix, `sentry-metrics` Deployment goes into ReplicaFailure because it cannot launch pods with non-existent ServiceAccount `sentry`.

Mokto:
Sure, thanks! Sorry I'm super busy these days.

PR #469, opened 2021-09-07
Implement dnsPolicy and dnsConfig

vaivanov:
Added dnsPolicy and dnsConfig properties in order to configure [DNS for pods](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/)

### DIFF
--- a/sentry/templates/cronjob-sentry-cleanup.yaml
+++ b/sentry/templates/cronjob-sentry-cleanup.yaml
@@ -40,6 +40,13 @@ spec:
           tolerations:
 {{ toYaml .Values.sentry.cleanup.tolerations | indent 12 }}
             {{- end }}
+            {{- if .Values.dnsPolicy }}
+          dnsPolicy: {{ .Values.dnsPolicy | quote }}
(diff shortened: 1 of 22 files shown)


Mokto:
Thanks

soer7022:
@vaivanov is it possible to configure a cluster domain suffix to use with these nameservers?
After adding these lines to my values.yaml I get the following in my `/etc/resolv.conf`
```
root@****-sentry-web-7bcbc779ff-s4dsb:/# cat /etc/resolv.conf 
nameserver 172.20.0.10
nameserver 10.148.***.***
nameserver 10.149.***.***
search default.svc.cluster.local svc.cluster.local cluster.local eu-west-1.compute.internal ***.***.com ***.***.com
options ndots:5
```
*** values censored for work reasons :)
This does not work since the 172.20.0.10 takes precedence over the custom servers so I cannot find internal sites...

soer7022:
It would be awesome if I could use my dns servers for urls that end with mycompany.com and use the 172.20.0.10 for the rest

PR #474, opened 2021-09-09
Fix snuba cleanup env

maths22:
Currently if you define a custom env for the cleanup jobs, it fails because the
indentation doesn't line up.  We might as well just use the same env as for the
other snuba pods.﻿


maths22:
Looking back this is a partial dupe of #470 . That one fixes using overrides but doesn't unify the env configuration.  I think this is preferable but it's not a strong opinion.

Mokto:
As I already merged the other one, I think I'm going to close this one.

Thanks anyway!

PR #476, opened 2021-09-10
Add missing resources to templates and values.yaml

yshch:
Currently there is no way to deploy the chart into a cluster with enabled ResourceQuota admission control plugin. This PR adds a few missing `resources:` keys to the templates and values.yaml.

### DIFF
--- a/sentry/templates/deployment-relay.yaml
+++ b/sentry/templates/deployment-relay.yaml
@@ -86,6 +86,8 @@ spec:
       dnsConfig:
 {{ toYaml .Values.dnsConfig | indent 8 }}
       {{- end }}
+          resources:
+{{ toYaml .Values.relay.init.resources | indent 12 }}
       containers:
       - name: {{ .Chart.Name }}-relay
         image: "{{ template "relay.image" . }}"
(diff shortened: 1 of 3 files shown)


prskr:
While I welcome this PR you could easily work around this with a [LimitRange](https://kubernetes.io/docs/concepts/policy/limit-range/) in the namespace you want to deploy this chart.

We've also enabled `ResourceQuota`s and haven't had any issues :blush:

Mokto (approved):

Mokto:
Can you rebase your branch please ? Thanks!

yshch:
@Mokto Done, sorry for the delay.

yshch:
> While I welcome this PR you could easily work around this with a LimitRange in the namespace you want to deploy this chart.

@baez90 Yeah, this is a good solution most of the time. I still think setting explicit requests and limits should be an option. Some cluster admins don't like to enable default requests, because without explicit errors the users may get complacent about setting the requests correctly. Also this one is not the only chart that doesn't allow explicit requests for init or test containers, so you can easily end up with a bunch of them in a single namespace and then you're stuck with default requests for a bunch of totally different workloads :slightly_smiling_face:

prskr:
@yshch sure thing! I didn't mean that 'LimitRange's replace your PR or something similar I just thought better mention it as a workaround for someone running into this problem until your PR is merged :blush: 
I'm looking forward to be able to actually assign limits for the init containers as well :blush:

Mokto:

PR #480, opened 2021-09-14
Support additional hostnames in Ingress

### LINKED ISSUE #479: Support multiple host names in Ingress
Just stumbled upon the need to support multiple host names.
An `additionalHostNames` variable for the `Ingress` resource would allow to define more than one host name without breaking the current chart.

E.g. the `nginx.ingress.kubernetes.io/upstream-vhost` allows nginx users to rewrite the `Host` header so that no CSRF issues should arise (at least I didn't notice any when I tested the setup in my test environment).

The TLS setup already supports multiple host names, hence nothing to change there.
I'll provide a PR to probably further discuss implementation details

prskr:
Closes #479


Mokto:
Thanks. Could you please bump the minor instead of the patch version?

prskr:
Sorry, version bump was actually not intended.
I just bumped it for internal testing and mistakenly pushed it but bumped it to 11.8.0 as requested.
Thanks!

Mokto:
You also have some weird binary files in your PR

prskr:
I'll remove them immediately - sorry for the extra effort :sweat_smile:

Mokto:
No worries & thanks!

Issue #492, opened 2021-09-27
Error: statefulsets.apps "sentry-clickhouse" is forbidden: 


cdesch:
Has anyone encountered this error before when deploying sentry? I was deploying manually with:

     helm install sentry sentry -n sentry -f sentry/values.yaml

Error Message:

     Error: statefulsets.apps "sentry-clickhouse" is forbidden: non-admin user "3432-2342-23423432-234234"
     [service account "sentry:default"]. The configured privileged attributes access for non-admin users
     ("[]")("[]") and for service accounts ("[]")("[]") lack required permissions to use attributes [privileged] 
     for resource sentry-clickhouse

sagor999:
clickhouse is deployed with:
```
securityContext:    
  privileged: true  
  runAsUser: 0      
```
which seems like your cluster is not allowing. I would check with whoever manages your cluster to see if they can allow clickhouse to get deployed with privileged permissions.

cdesch:
Why does Clickhouse need this priviedge? I can try working with the cluster admins on this, but if it is not something that is doable, can clickhouse be disabled?

sagor999:
That's a good question. 
It seems like there is no "official" helm chart for clickhouse. Sentry uses helm chart that seems to be cloned from this repo: https://github.com/liwenhe1993/charts/tree/master/clickhouse
which has not been updated for two years.

I am not sure if there is an actual reason for clickhouse to run with privileged access, but maybe @Mokto knows.

Mokto:
To be honest, no idea!

cdesch:
Thanks @Mokto @sagor999 !

I'm not familiar with clickhouse, so I could be wrong here, but I'm still not sure why clickhouse would need the elevated privileges. I tried forking the [clickhouse helmchart](https://github.com/liwenhe1993/charts) and commenting out: 

```
securityContext:    
  privileged: true  
  runAsUser: 0      
```

and things appear to work fine on my local machine, but that doesn't take into context if `sentry-kubernetes` has a requirement for clickhouse to have the elevated permissions.

[...]

PR #493, opened 2021-09-28
Remove clickhouse securityContext & bump to v3.1.0

cdesch:


cdesch:
this PR is related to the issue [492](https://github.com/sentry-kubernetes/charts/issues/492)

prskr:
Why not making the `securityContext` configurable from the values file to allow others to add some options if they need/want?

cdesch:
Thanks @baez90 ! I made the Security Context Configurable. Could @baez90 please review my changes? I'm new to working with helm templates and would like any feedback or review you can provide on the changes.

clickhouse/values.yaml
```diff
@@ -48,7 +48,13 @@ clickhouse:
   ## https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions
   ##
   # rollingUpdatePartition:
-
+  
+  # Security Context
+  securityContext:
+    enabled: false
```
prskr: nitpick: I would get rid of this `enabled` property just by either setting the `securityContext` to something like you did or by setting it empty like so:

```yaml
securityContext: {}
````

File: clickhouse/templates/statefulset-clickhouse-replica.yaml [hunk 2]
```diff
@@ -116,9 +116,11 @@ spec:
           mountPath: /etc/clickhouse-server/metrica.d
         - name: {{ include "clickhouse.fullname" . }}-users
           mountPath: /etc/clickhouse-server/users.d
+        {{- if .Values.clickhouse.securityContext.enabled }}
```
prskr: continuation:

assuming you removed the `enabled` property you can make this conditional like so:

```yaml
{{- with .Values.clickhouse.securityContext }}
        securityContext:
          {{- toYaml . | nindent 10 }}
{{- end }}
``` 

that way the `securityContext` will only be included if the `.Values.clickhouse.securityContext` is not nil or empty

this way the content of the `securityContext` is also flexible because it can be set from the `values.yaml` without requiring a property for every possible field to be set

clickhouse/templates/statefulset-clickhouse.yaml
```diff
@@ -115,9 +115,11 @@ spec:
           mountPath: /etc/clickhouse-server/metrica.d
         - name: {{ include "clickhouse.fullname" . }}-users
           mountPath: /etc/clickhouse-server/users.d
+        {{- if .Values.clickhouse.securityContext.enabled }}
```
prskr: of course the same applies here :blush:

prskr (commented):
Just an idea that might be interesting for you if you're new at Helm templating. As I am no maintainer in this repository I can't approve anything here :smile:

clickhouse/values.yaml
cdesch: thanks @baez90 . I like this idea better, thanks for the feedback. I was initially basing it off what I saw with `metrics` section in `values.yaml`

```
metrics:
    enabled: false
    image:
      registry: docker.io
      repository: f1yegor/clickhouse-exporter
      tag: latest
      pullPolicy: IfNotPresent
      port: 9116
```

Would it be better to have the `enabled` property if there were default values that we wanted to provided with the `securityContext`?

clickhouse/templates/statefulset-clickhouse-replica.yaml
cdesch: @baez90 Thanks again! yeah, this looks better for customizing the `securityContext`

clickhouse/templates/statefulset-clickhouse.yaml
cdesch: Yes. Thanks @baez90 . I made all of the updates. Thank you again for the feedback!

clickhouse/values.yaml
prskr: Not sure about 'better' but as a rule of thumb I use: when I want to exclude a whole resource I prefer`enabled` but thinking about your comment you're absolutely right that it makes sense always when you have default values you'd like to include in the values file and you don't want to comment them for instance.
I don't think there's a hard rule more personal preferences :blush:

cdesch: I agree. Thanks again!

cdesch:
@Mokto Is there anything else we should be considering with this PR?

Mokto (approved):

Mokto:
No it's fine, thanks!

PR #494, opened 2021-09-29
feat: smtp support use-ssl options

anjia0532:
[TLS in SMTP is STARTTLS that is less secure than SSL(TLS) #22492](https://github.com/getsentry/sentry/issues/22492)


Mokto (approved):

Mokto:
Thanks!

PR #510, opened 2021-10-22
fix: use Snuba migrations to generate distributed tables

J0sh0nat0r:
This PR enables distributed table generation via the Shuba jobs, rather than our own. This fixes an error observed in Sentry 21.10 (a table now has a `TTL`, which prevents us using an `AS` expression since `TTL`s are not supported with the `Distributed` engine) as well as future-proofing us against any similar breaking changes.

Note: I have not tested whether the distributed tables will need to be manually dropped on existing installations, however should it prove necessary it will be possible without data loss.


Mokto:
I created this PR that includes the october release & your code: https://github.com/sentry-kubernetes/charts/pull/521

Everything seems to work fine for me except I still have issues with the performances tab.

prskr:
I experienced the same issue @Mokto has with the performance tab and with this fix and 'some' extra steps I could resolve it.
What I did:

* migrate from the ClickHouse chart to the ClickHouse operator and initialize a new ClickHouse cluster ( version `docker.io/yandex/clickhouse-server:20.3.9.70` just like Sentry is doing it right now) - this is not mandatory, you could do it also in place (in fact I tested it in place)
* create a backup of the currently running ClickHouse database e.g. with [this tool](https://github.com/AlexAkulov/clickhouse-backup) (make sure to run it as a sidecar with the exact same volume mounts as ClickHouse to get the best experience)
* run `snuba migrations migrate --force` with the new ClickHouse instance/drop all tables of your instance previosly (this makes sure the expected tables are created especially the distributed ones which I had the most problems with)
* now things are getting a bit messy (assuming you used the aforementioned backup tool) because you can't just restore the backup but you've to rename the data directories and remove all obsolete table definitions. I'll update this as soon as I applied this procedure to my prod system with exact steps but I hope you get the idea.

I'm currently running Sentry 21.11.0 in testing without a problem/any error I found so far

**Edit:** I just realized that I also configured my ClickHouse instance with ZooKeeper which is a dependency for the new database scheme because Sentry's using `ReplicatedReplacingMergeTree` as engine now

PR #537, opened 2021-12-22
Fix healthchecks under AWS ALB

bdwyertech:
This fixes the perpetually failing healthchecks under AWS ALB.  This sets the appropriate URI Path to the services respective healthcheck endpoint and explicitly sets the appropriate port.


Mokto (approved):

sfc-gh-pkommini:
@bdwyertech @Mokto Per [this](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#subnets) documentation, the `aws-load-balancer-controller` seems to have different annotations for kubernetes `Service` object.

So these aren't really working. I'm now trying to pass the service annotations. Will report once I've tried it out.

Issue #532, opened 2021-12-07
Kafka event queue keeps getting out of Sync since v12


colincoghill:
While investigating why our Kafka event queue lag suddenly starts growing and never comes down, resulting in the kinds of problems discussed in several other issues, I find:

sentry-snuba-subscription-consumer-events
sentry-snuba-subscription-consumer-transactions

both attempt to follow the same queue:   
         Partition(topic=Topic(name='events'), index=0)
but since there's only one partition in our (default?) setup, only one can do this at any given time, and whichever
loses a race gets:

'''Partitions revoked: [Partition(topic=Topic(name='events'), index=0)]'''

and no longer takes events from the queue, causing it to build up.

The two lines of code that set this up are:

https://github.com/sentry-kubernetes/charts/blob/3b6591d88244808a894bc0cfd32f257fcbe67a1c/sentry/templates/deployment-snuba-subscription-consumer-transactions.yaml#L73

and

https://github.com/sentry-kubernetes/charts/blob/3b6591d88244808a894bc0cfd32f257fcbe67a1c/sentry/templates/deployment-snuba-subscription-consumer-events.yaml#L73

Am I on the right track? Should they be using different queues, or is there some way I can add more partitions, or is there something else I can do to solve this?

vaivanov:
Looks like a bug - in [the original docker-compose file](https://github.com/getsentry/self-hosted/blob/master/docker-compose.yml#L250) they used for "snuba-subscription-consumer-transactions" and "snuba-events-subscriptions-consumers" consumer groups instead of use the same ("snuba-events-subscriptions-consumers")

colincoghill:
Yeah, that seemed to fix it for me:

templates/deployment-snuba-subscription-consumer-transactions.yaml:

```
72c72
<           - "--consumer-group=snuba-events-subscriptions-consumers"
---
>           - "--consumer-group=snuba-transactions-subscriptions-consumers"
74c74
<           - "--result-topic=events-subscription-results"
---
>           - "--result-topic=transactions-subscription-results"
```

cabrinha:
So whats the fix for this?

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #543, opened 2022-01-07
Fix: issue-532. Snuba using incorrect kafka topic/group.

colincoghill:
I believe snuba was using the wrong queue for transactions subscription, making it conflict with the events queue, causing things to get blocked up.  

I've set it to use what I believe is the right one. This seems to work well for me.

This is my first sentry PR, I think this is right, but can someone verify this is correct?   :)

Let me know if I need to do anything else.

https://github.com/sentry-kubernetes/charts/issues/532


z0rc:
@Mokto Any comments on this? This is quite important fix and it wasn't included in 13.0.0 for some reason.

Mokto:
Thanks. Pushing it as 13.0.1

PR #552, opened 2022-01-18
Add urlquery method to redis configuration of relay service

tttlkkkl:
When I use external redis and my password contains the '@' character, the relay service cannot resolve to the correct redis host. An error will be reported as follows:
```bash
 [r2d2] ERROR: failed to lookup address information: Name or service not known
```
Therefore, add the 'urlquery' method to the password field for escape processing.


Mokto:

PR #560, opened 2022-01-25
Relay Health-check liveness endpoint added

The-Seyed:
Accroding to Sentry relay [documentation](https://docs.sentry.io/product/relay/monitoring/#health-checks), there are two endpoints provided for relay health-check. Only .../ready endpoint was being called in both K8S Liveness probe and Readiness probe. So, why not use both provided endpoints...!


Mokto (approved):

Mokto:
Could you please rebase your branch ? Thanks!

PR #570, opened 2022-02-09
Add linting to enforce bumping the version on changes to charts

### LINKED ISSUE #456: [Sentry] Enforce Chart version bump at CI
When using the Sentry chart, we will often make incremental tuning changes to our Sentry deployment based on the workload that comes in (autoscaling counts, pod settings, etc.). Sometimes, we will run our deploy job which diffs the version deployed to our cluster with the new version, and we will see changes apart from our own even when we have not changed the version!

Almost all of the time this is fine and works great for us, as the new changes are usually improvements. That being said, it does make us wonder what happens if a bad change comes through and we are negligent about reviewing differences between the deployed version and the new version. We could potentially get Sentry into a bad state.

With that disaster scenario in mind (potentially deploying a new version of the chart when there has been no versioned change), I was wondering if it would be possible to add a CI time check to enforce version bumps? That way surprise changes like https://github.com/sentry-kubernetes/charts/pull/446 don't get deployed on the same version and users have to opt in to the changes by going up a version (patch, minor, etc.).

Happy to submit a PR for this, but I wanted to know if this is wanted by the community first (or if another CI process would be too onerous).

jandersen-plaid:
This fixes #456 where sometimes pr authors would forget to bump the Chart.yaml version. While safe most of the time, as authors made backwards-compatible fixes in most cases, this was dangerous as it unintentionally created a different Chart for the same version.

This PR adds in the `chart-testing` github action to check if the `Chart.yaml` has been changed on editing the chart.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -20,14 +20,14 @@ jobs:
       - name: Run chart-testing (list-changed)
         id: list-changed
         run: |
-          changed=$(ct list-changed --config ct.yaml)
+          changed=$(ct list-changed --target-branch develop --chart-dirs . )
           if [[ -n "$changed" ]]; then
             echo "::set-output name=changed::true"
           fi
       - name: Run chart-testing (lint)
-        run: ct lint --config ct.yaml
+        run: ct lint --target-branch develop --chart-dirs .
       - name: Create kind cluster
         uses: helm/kind-action@v1.2.0
         if: steps.list-changed.outputs.changed == 'true'
       - name: Run chart-testing (install)
-        run: ct install --config ct.yaml
+        run: ct install --target-branch develop --chart-dirs 
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -20,12 +20,10 @@ jobs:
           fi
       - name: Lint Sentry Helm Chart
         uses: WyriHaximus/github-action-helm3@v2
-        needs: list-changed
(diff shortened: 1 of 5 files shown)


jandersen-plaid:
Finally got around to implementing this @Mokto -- let me know if there are any additional changes that you would like to be made.

Mokto:
Thanks! 

1. Why do you use depth = 0 ? 
2. Maybe you could apply this to clickhouse too?

jandersen-plaid:
> 1. Why do you use depth = 0 ?

The Github Actions checkout does a shallow checkout by default, which breaks chart-testing since git can't perform a merge-base that is requested of it in `ct list-changed` ([ref](https://github.com/helm/chart-testing/issues/186#issuecomment-602052815)).

Since I was testing this on my fork, the failing action isn't visible here, but you can [check there for the failing action for the exact failure mode (exit code 128)](https://github.com/jandersen-plaid/charts/runs/5126033618?check_suite_focus=true).

> 2. Maybe you could apply this to clickhouse too?

Absolutely!

jandersen-plaid:
Alright, everything should be working now. [cc3ddfb](https://github.com/sentry-kubernetes/charts/pull/570/commits/cc3ddfb00bd482dd45d13c30bfb8e628917fddec) should have validated that it works as intended (though clickhouse is failing linting because the `apiVersion` is required in the `Chart.yaml`)

Mokto:
Thanks!

PR #576, opened 2022-02-22
Correct Indentation Level - Db Init Job

Silvenga:
When using a env var with the sentry-db-init job, helm throws the error:

```
Error: YAML parse error on sentry/templates/hooks/sentry-db-init.job.yaml: error converting YAML to JSON: yaml: line 37: did not find expected key
```

This pr resolves this problem.


Mokto:
Thanks!

PR #581, opened 2022-03-09
Do not mount any volume if filestore backend is not 'filesystem'

prskr:
Currently it is not really possible to use the S3 filestore backend because the chart always wants to mount a data volume even though it's not necessary.

This PR changes the behavior to only mount the `sentry-data` volume if `filestore.backend == "filesystem"`


revathi-murali:
@baez90 Any reason why this PR is closed? This change is required in my opinion

prskr:
I cleaned up my open PRs and as this one didn't get any feedback for quite some while I figured it isn't required :shrug:

PR #582, opened 2022-03-11
deploy replica services only when replica is deployed

foosinn:
this removes all clickhouse-replica services if the replica statefulset is not deployed.

we monitor services with no endpoints, this would prevent deploying services that are neither used nor have any targets.


Mokto:
Thanks for your PR.

I'm a bit confused about your usage of the `backup` property?

foosinn:
Hey,

tbh i just copied it from the matching statefulset:

https://github.com/sentry-kubernetes/charts/blob/develop/clickhouse/templates/statefulset-clickhouse-replica.yaml#L1

Mokto:
Fair enough, thanks ;)

PR #584, opened 2022-03-11
Pass external postgres password to cleanup job

wacuuu:
I noticed that external postgres password is passed to sentry configmap, but not to sentry cleanup job. This PR adds this possibility if the password is set in values file


Mokto:
Thanks!

PR #588, opened 2022-03-23
Feat: distributed tables v2

J0sh0nat0r:
This PR incorporates changes and suggestion in comments from #521 in-order to support the latest version of Sentry.

This necessitates a major version bump, so is a good oppertunity to get any more bc breaking changes in.


Mokto:
Wow, amazing! Thanks :) 

Just tried it and it seems to work flawlessly. Giving it a couple more hours but then we can merge it & upgrade to the latest Sentry version.

sentry/values.yaml
```diff
@@ -699,12 +703,20 @@ externalClickhouse:
...
+  singleNode: true
   ## Cluster name, can be found in config
   ## (https://clickhouse.tech/docs/en/operations/server-configuration-parameters/settings/#server-settings-remote-servers)
   ## or by executing `select * from system.clusters`
   ##
   # clusterName: test_shard_localhost
 
+# Settings for Zookeeper.
+# See https://github.com/bitnami/charts/tree/master/bitnami/zookeeper
+zookeeper:
```
danilutkin, line 715: Zookeeper is setting up by kafka chart. 
This part with zookeeper config just confusing, in fact only zookeeper from  kafka chart is working.

PR #589, opened 2022-03-25
Fix deprecated api`s version batch/v1beta1

alexvivanov:
Helm 3.8 lint says: batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob


Mokto:
Fixed with 14.0.0

Issue #67, opened 2020-05-15
OpenShift unable to deploy because of privileged users


haarchri:
hello,

it would be great to ship all statefulsets that it is possible to run in openshift 

for example:
`create Pod sentry-clickhouse-replica-0 in StatefulSet sentry-clickhouse-replica failed error: pods "sentry-clickhouse-replica-0" is forbidden: unable to validate against any security context constraint: [spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 0: must be in the ranges: [1016890000, 1016899999] spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]`

DandyDeveloper:
@haarchri Unfortunately this is actually a dependency issue that would need to be fixed by Clickhouse itself: 

https://github.com/liwenhe1993/charts/blob/master/clickhouse/templates/statefulset-clickhouse.yaml#L94

You can raise an issue there and we may be able to contribute a solution. 

In the meantime, you should be able to configure OpenShift to allow this (If your organisation permits).

Mokto:
Because of this, [this other issue](https://github.com/sentry-kubernetes/charts/pull/50#issuecomment-612444903) and the lack of maintenance, I'm more and more thinking of building a fork of clickhouse inside that repository.

What do you think?

DandyDeveloper:
@Mokto Wouldn't hurt, but it's another thing to manage ;) That's the only downside.

I'm not the most confident with clickhouse, but I'm sure it can't be that hard to follow.

haarchri:
the only open point is clickhouse in the helm chart - i play around with this implementation https://github.com/Uglykoyote/clickhouse will have a look if its possible with this openshift template - if yes i can create a helm chart of it 

the rest is starting up with a values.yaml like this:

```
prefix:

user:
  create: true
  email: test@example.com
  password: aaaa

images:
  sentry:
    repository: getsentry/sentry
    tag: 6e6ed914e24d9121619ccc754325fbba611ad6be
    pullPolicy: IfNotPresent
    # imagePullSecrets: []
  snuba:
    repository: getsentry/snuba
    tag: fdcc3229b24b659a48b4934e4ffc0ccaa6364a37
    pullPolicy: IfNotPresent
    # imagePullSecrets: []

sentry:
  web:
    replicas: 1
    env: {}
    probeInitialDelaySeconds: 10
    resources: {}
    affinity: {}
    nodeSelector: {}
[...]

PR #609, opened 2022-04-15
feat: add value to enable/disable init of ClickHouse

minijus:
When running in restricted OpenShift environment we were getting this error from `initContainer`:
```
mkdir: can't create directory '/etc/clickhouse-server/': Permission denied
```

Trying to remove the `initContainer` completely yielded in no error and the pods started successfully. It seem that the `initContainer` is not needed (or at least no longer needed for some later version of ClickHouse).


Mokto:
Hey, thanks for your PR.

What is the purpose of it?

minijus:
Hi @Mokto,

I've updated description of this PR to include some details and reasoning why I did it.

pauvos:
Really good, can we merge this please?

The busybox init container doesn't work with PodSecurityPolicies (runAsUser > 0) enabled. I can confirm that clickhouse starts perfectly without it - at least a recent [clickhouse/clickhouse-server](https://hub.docker.com/r/clickhouse/clickhouse-server) version.

It appears the yandex repo is deprecated? https://github.com/sentry-kubernetes/charts/issues/67#issuecomment-1104118407

Mokto:
Hey guys, thanks for your PR.

If we don't need it anymore, why don't we remove the container entirely?

pauvos:
@Mokto I think you are right and it can be dropped. Since there is no volume mounted to `/etc/` it probably doesn't do anything.

Concering clickhouse: the latest clickhouse/clickhouse-server version (22.x) starts but doesn't seem to work with sentry (missing  schema migrations not running). I found a workaround for yandex/clickhouse-server and will leave a comment in #67 .

Mokto:

PR #631, opened 2022-05-28
Allow install clickhouse with external zookeeper

z0rc:
Helm evaluates conditions by first seen instead of logical AND. If user wants to install clickhouse with external zookeeper, it wasn't possible to disable zookeeper installation by setting `zookeeper.enabled: false`, because `clickhouse.enabled: true` was evaluated first.

By dropping `clickhouse.enabled` from condition, zookeeper can be disabled as expected by setting `zookeeper.enabled: false`.

Backward compatibility is preserved, as default values already define `zookeeper.enabled: true`.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -20,7 +20,7 @@ dependencies:
   - name: zookeeper
     repository: https://charts.bitnami.com/bitnami
     version: 9.0.0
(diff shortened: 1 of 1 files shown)


z0rc:
@Mokto I decided to keep PR changes to be functional only. Feel free to version and release them as you see fit.

z0rc:
@Mokto ping

PR #632, opened 2022-05-29
Update symbolicator to 0.5.0

z0rc:
It's correct version that should be used with sentry 22.5.0, see https://github.com/getsentry/self-hosted/blob/22.5.0/.env#L11.

### DIFF
--- a/sentry/values.yaml
+++ b/sentry/values.yaml
@@ -29,7 +29,7 @@ images:
     imagePullSecrets: []
   symbolicator:
     # repository: getsentry/symbolicator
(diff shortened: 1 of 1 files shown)


z0rc:
@Mokto what's the process related to version bump? I have 2 PRs open and one more incoming. How at this case should the version change be handled? If I change version in each PR, that would require changing PRs again after one of them is merged. Is this right and expected?

Mokto:
That's ok and expected.

You can also leave the version untouched if you have incoming PRs.

Your choice ;)

PR #574, opened 2022-02-18
FIX metrics alert not firing

marcomezzaro:
Pull request as requested for #388

PR #634, opened 2022-06-03
(fix) Update the snuba transactions consumer to use a commit log topic

jandersen-plaid:
This PR brings the commands used for snuba-transactions-consumer up to date with the sentry on-premise docker-compose commands: https://github.com/getsentry/self-hosted/blob/master/docker-compose.yml#L247


jandersen-plaid:
Ah, I forgot to check the open PRs -- https://github.com/sentry-kubernetes/charts/pull/574 should be approved instead of this.

PR #639, opened 2022-06-13
[sentry] Bitnami dep bump

### LINKED ISSUE #637: Error: can't get a valid version for repositories redis, kafka, rabbitmq, postgresql, nginx.
I'm getting the following error:
`Error: can't get a valid version for repositories redis, kafka, rabbitmq, postgresql, nginx. Try changing the version constraint in Chart.yaml`

The problem was fixed as soon as I updated the versions in Chart.yaml:
```
apiVersion: v2
name: sentry
type: application
version: 14.1.1
appVersion: 22.5.0
dependencies:
  - name: redis
    repository: https://charts.bitnami.com/bitnami
    version: 15.7.6
    condition: redis.enabled
  - name: kafka
    repository: https://charts.bitnami.com/bitnami
    version: 14.9.3
    condition: kafka.enabled
  - name: clickhouse
    repository: https://sentry-kubernetes.github.io/charts
    version: 3.1.2
    condition: clickhouse.enabled
  - name: zookeeper
    repository: https://charts.bitnami.com/bitnami
    version: 9.0.0
    condition: clickhouse.enabled,zookeeper.enabled
  - name: rabbitmq
    repository: https://charts.bitnami.com/bitnami
    version: 8.32.2
    alias: rabbitmq
    condition: rabbitmq.enabled
  - name: postgresql
    repository: https://charts.bitnami.com/bitnami
    version: 10.14.4
    condition: postgresql.enabled
  - name: nginx
    repository: https://charts.bitnami.com/bitnami
    version: 9.6.2
    condition: nginx.enabled
maintainers:
  - name: sentry-kubernetes
```

czomo:
Resolves https://github.com/sentry-kubernetes/charts/issues/637
Huge dependency bump because of https://github.com/bitnami/charts/issues/10539


czomo:
@Mokto  Can you take a look? Should we bump it to 15.x.x version? Do not merge it at this point. I will conduct additional checks

Mokto:
We should definitely bump it to 15.x.x.

Have you tried the upgrade with all subcharts?

Seems super risky. Zookeeper is upgraded to v3.8.0 for example.

czomo:
I will do my best. Will try to update it from `13.1.0` but first i need to find out how to add external zookeeper as the only way is to use the build one( see [pull](https://github.com/sentry-kubernetes/charts/pull/588), I am using AWS MKS.  And you right we need to keep Zookeeper in the same version as kafka. I will patch it along with bump of the chart

czomo:
Ok, I bumped version of the chart to 15.0.0 and fixed ingress template.
I followed update from `13.1.0` to `14.1.2` that went smooth. Then upgraded it using chart from my branch and it seems to be working. However I used external kafka(kafka/zookeper from AWS MSK are coupled together) and postgres. Will check now how it will work with parts together and keep you posted.

>Seems super risky. Zookeeper is upgraded to v3.8.0 for example

This drives me nuts. Version of kafka chart that we are using in 14.1.2 sould use zookeeper in version `6.x.x` of chart, [see](https://github.com/bitnami/charts/commit/32625a9f26a70f3aff20f8c5aebd7a8ec3e75f54). I donno why there is already in version [9.0.0](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/Chart.yaml#L22) In my opinion we should align in somewhat what bitnami did. Wdyt? @Mokto

czomo:
Found something strange. Kafka chart already have its zookeeper dependency, [see](https://github.com/bitnami/charts/blob/32625a9f26a70f3aff20f8c5aebd7a8ec3e75f54/bitnami/kafka/Chart.yaml#L6) I see no point to have it separately. Zookeeper is enabled in kafka chart by default and we could skip problems with mismatched dependencies. Take a look

J0sh0nat0r:
> Found something strange. Kafka chart already have its zookeeper dependency, [see](https://github.com/bitnami/charts/blob/32625a9f26a70f3aff20f8c5aebd7a8ec3e75f54/bitnami/kafka/Chart.yaml#L6) I see no point to have it separately. Zookeeper is enabled in kafka chart by default and we could skip problems with mismatched dependencies. Take a look

This charts Zookeeper dependency is actually used for ClickHouse:
https://github.com/sentry-kubernetes/charts/blob/5f55d8a07d6932ad405a90c6a35d588fbc04311c/sentry/values.yaml#L728-L730
https://github.com/sentry-kubernetes/charts/blob/5f55d8a07d6932ad405a90c6a35d588fbc04311c/sentry/values.yaml#L675-L679
An alternative might be pushing it down into the ClickHouse chart?

czomo:
Yes, it is used by clickhouse and [kafka require zookeeper too](http://cloudurable.com/blog/kafka-architecture/index.html). The clue is to have shared, high available zookeeper using included dependency from the kafka chart. Right now we are deploying two instances of zookeeper, one for kafka, second 3-replica for clickhouse. Check this out. Deploy `kafka` and `zookeeper` chart using standard values form sentry develop branch
```
 kgp
NAME                                      READY   STATUS              RESTARTS   AGE
chart-1655533724-kafka-0                  0/1     ContainerCreating   0          73s
chart-1655533724-kafka-1                  0/1     ContainerCreating   0          72s
chart-1655533724-kafka-2                  0/1     ContainerCreating   0          72s
chart-1655533724-zookeeper-0              0/1     ContainerCreating   0          73s  >>>kafka chart dependency
chart-1655533724-zookeeper-clickhouse-0   0/1     ContainerCreating   0          72s   >>> zookeeper chart dependency
chart-1655533724-zookeeper-clickhouse-1   0/1     ContainerCreating   0          72s    >>> zookeeper chart dependency
chart-1655533724-zookeeper-clickhouse-2   0/1     ContainerCreating   0          72s     >>> zookeeper chart dependency
```
This causes that kafka is usually unstable and we need to care about another dependency. That's why i moved zookeeper values to kafka https://github.com/sentry-kubernetes/charts/pull/639/files#diff-905e10a3c65b9d63e408c1a57d53f1793e6c3d8357e8a2fd25a5b05f824d23a1R746. User should be able to reuse pvc from zookeper chart without much hassle. Wdyt? @Mokto @J0sh0nat0r

J0sh0nat0r:
How about using the single Zookeeper dependency of this chart for both? I propose this because:

1. It makes migration easier (we can simply switch the Zookeeper instance Kafka is using with minimal or no data loss)
2. Kafka plans to eventually remove the need for Zookeeper with KIP-500

It seems like all we'd need to do is configure `externalZookeeper` and `zookeeperChrootPath`

Am I missing anything? If not that sounds resonable to me, unless @Mokto has anything

Mokto:
See https://docs.altinity.com/operationsguide/clickhouse-zookeeper/zookeeper-installation/#precautions

```
Do not share ZooKeeper with other applications like Kafka.
```

Not sure it would work tbh

z0rc:
I cannot say about generic use case of sharing zookeeper between kafka and clickhouse. But in case of AWS MSK, this works okay for me. AWS MSK is managed kafka, its zookeeper endpoints are accessible and clickhouse works fine with these endpoints. AFAIK it's just a matter of configuring kafka and clickhouse to use separate znodes/roots.

UPD I did small research, by default clickhouse puts znodes under `/clickhouse` root, which doesn't conflict with kafka, even when kafka uses `/` root.

UPD2 AFAIU running single zookeeper for multiple applications is okay, if you understand what you're doing, run it in HA mode, and know how to handle single node failures. All of which more or less covered by managed zookeeper from AWS MSK. But I don't know how to answer on these questions for zookeeper deployed via chart.

czomo:
I share the same experience experience as @z0rc using shared  zookeeper via AWS MSK, it works just fine, clickhouse puts everything under `clickhouse` path as specified in [here](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L678)

Let's make it more clear, update dependencies under this PR and then think about shared zookeeper. I will revert removal of zeekeeper.

czomo:
@Mokto ping

Mokto:
Just tried it, seems to work!

Thanks.

PR #645, opened 2022-06-14
Clickhouse and podsecuritypolicy

pauvos:
my approach to solve the PodSecurityPolicies issues described in #67:
* remove the existing init container (#609)
* introduction of `clickhouse.command` helm value so ppl can use their own hacks to disable the directory-ownership-check present in the entrypoint script of <21.2 clickhouse images
* add `clickhouse.initContainers` in case anyone needs it

### DIFF
--- a/clickhouse/templates/statefulset-clickhouse-replica.yaml
+++ b/clickhouse/templates/statefulset-clickhouse-replica.yaml
@@ -74,9 +74,7 @@ spec:
         image: {{ .Values.clickhouse.image }}:{{ .Values.clickhouse.imageVersion }}
         imagePullPolicy: {{ .Values.clickhouse.imagePullPolicy }}
         command:
-          - /bin/bash
-          - -c
-          - export SHARD=${HOSTNAME##*-} && /entrypoint.sh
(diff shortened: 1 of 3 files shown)


Mokto:
Could you rebase your branch please? I would like to try it on my own environment.

pauvos:
i rebased the changes to sentry-kubernetes:develop. should i also bump the chart version?

Agalin:
There is an alternative solution which doesn't require end user to hack the command - instead of mounting `/var/lib/clickhouse` and storing data in the same path use `/var/lib/clickhouse/data` (or any other subdirectory) - entrypoint tries to create this path and (at least in Kubernetes) it gets correct permissions. It would require separation of `clickhouse.fullpath` (constructed from `.Values.clickhouse.path`) into two variables - one for the mount point and another for data path.
BTW same issue happens with `/var/log/clickhouse-server` directory (`clickhouse.logpath` based on `.Values.clickhouse.configmap.logger.path`) - in this case IMO whole `/var/log` can be mounted as emptydir.

Mokto:

Issue #647, opened 2022-06-16
Snuba migrate job should be fixed for Sentry 22.6.0


andrdoc:
https://github.com/sentry-kubernetes/charts/blob/14.1.0/sentry/templates/configmap-snuba.yaml#L30-L46
missing 2 new sets from
https://github.com/getsentry/snuba/blob/22.6.0/snuba/settings/settings_distributed.py#L27-L28

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #649, opened 2022-06-17
Sentry 22.6.0

z0rc:
Ref:
- https://github.com/sentry-kubernetes/charts/issues/647
- https://github.com/getsentry/self-hosted/pull/1507
- https://github.com/getsentry/self-hosted/commit/eaed29a058c3b7769cc777cd8ef1100cec5b80d7

@Mokto I didn't update chart version, so there won't be a conflict with other PRs, like #639 and #581.

Also I would appreciate some additional testing, as my deployment is quite specific where only sentry and clickhouse are deployed via chart, rest of dependencies are externalized.

### DIFF
--- a/sentry/templates/deployment-snuba-subscription-consumer-events.yaml
+++ b/sentry/templates/deployment-snuba-subscription-consumer-events.yaml
@@ -66,18 +66,17 @@ spec:
         imagePullPolicy: {{ default "IfNotPresent" .Values.images.snuba.pullPolicy }}
         command:
           - "snuba"
-          - "subscriptions"
+          - "subscriptions-scheduler-executor"
           - "--partitions={{ .Values.snuba.subscriptionConsumerEvents.partitions}}"
-          - "--auto-offset-reset"
-          - "{{ .Values.snuba.subscriptionConsumerEvents.autoOffsetReset }}"
-          - "--consumer-group=snuba-events-subscriptions-consumers"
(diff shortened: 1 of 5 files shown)


Mokto:
Trying now, thanks!

sentry/templates/deployment-snuba-subscription-consumer-events.yaml
```diff
@@ -66,18 +66,17 @@ spec:
         imagePullPolicy: {{ default "IfNotPresent" .Values.images.snuba.pullPolicy }}
         command:
           - "snuba"
-          - "subscriptions"
+          - "subscriptions-scheduler-executor"
           - "--partitions={{ .Values.snuba.subscriptionConsumerEvents.partitions}}"
```
z0rc: It seems this isn't needed with `subscriptions-scheduler-executor`, at least this option isn't available anymore, see https://github.com/getsentry/snuba/blob/22.6.0/snuba/cli/subscriptions_scheduler_executor.py vs https://github.com/getsentry/snuba/blob/22.6.0/snuba/cli/subscriptions.py.

Might be worth removing it in next release.

z0rc: Yeah, `--partitions` option isn't available with new command. Removed it.

Mokto:
Thanks!

PR #659, opened 2022-07-06
feat: add podLables to Sentry hooks

nishaero:
The init jobs do not have unique podLables
The lables can be used for mapping these init jobs with services
Use case: To allow access to services inside
Consul mesh, a service is needed.


sentry/values.yaml
```diff
@@ -456,6 +459,8 @@ hooks:
     affinity: {}
     nodeSelector: {}
     # tolerations: []
+  snubaMigrate:
```
Mokto: Don't you need to use `snubaMigrate: {}`?

nishaero: > Don't you need to use `snubaMigrate: {}`?

Yes, I have added it. Thank You :)

Mokto (approved):

sentry/values.yaml
nishaero: I have bumped the chartversion to 15.0.1 this should fix the lint issue

Mokto (approved):

Issue #666, opened 2022-07-17
sentry-relay init container generates new credentials on each start


pulse00:
The init container from the sentry-relay pod is generating a new relay credentials.json file on each run. From how in understand the logic of this credentials file, it is not how it's meant to be used (see https://docs.sentry.io/product/relay/getting-started/#registering-relay-with-sentry).

Since it generates new credentials on each run, after registering the relay with the sentry web ui, it won't be able to create a connection after the next start, since it will have a new public/private key.

Maybe the sentry-relay `generate credentials` should better be implemented as a helm hook instead of an init container of the pod?

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #669, opened 2022-07-17
added sentry-relay-init job to only generate relay credentials once after installation

pulse00:
This PR removes the init-container from the relay pod and replaces the logic with a post-install hook to only generate the credentials once after installation.

References https://github.com/sentry-kubernetes/charts/issues/666


Mokto:
Thanks for your PR. How would that work for new installations?

Mokto:
Even for my current installation, my relay pods get stuck saying 
```
2022-07-18T05:12:03Z [relay_log::utils] ERROR: relay has no credentials, which are required in managed mode. Generate some with "relay credentials generate" first.
```

pulse00:
> Thanks for your PR. How would that work for new installations?

good point - unfortunately `relay credentials generate` will exit with an error if the credentials already exist and i found no option to exit with a different error code. 

Maybe we should override the command with a custom entry point script, check if the credentials.json already exists and only run the built in entry-point if it does not exist? then it could run as post-install and post-upgrade hook.

pulse00:
@Mokto i pushed a change that creates a configmap with a script which will check if the credentials are already present. it will do nothing if they are, but will generate them if they are not present.

Mokto:
I don't think this scales properly, you won't be able to mount the PVC to more than 1 container.

pulse00:
> I don't think this scales properly, you won't be able to mount the PVC to more than 1 container.

you're right. i refactored the PR to not use a PVC and be backwards compatible. It's possible to provide `relay.existingSecret` with a pre-generated credentials.json - if that exists, the initPod will not generate new credentials but copy over the existing ones from the secret.

Danielss89:
+1 for this one. Atm. we need to "register" the relay again and again if relay pod is deleted, as it will create new credentials on each new deploy

nklmilojevic:
Following up on this one. Is there a possibility for someone to review and merge it?

Mokto:

PR #641, opened 2022-06-13
allow using externalPostgresql.existingSecret to inject password

orezi:
Addresses #613 to add an existing secret that holds postgresql-password

PR #672, opened 2022-07-18
fixed re-generation of secret key

pulse00:
This PR uses the [environment variable](https://github.com/thinkingmachines/sentry-onpremise/blob/master/sentry.conf.py#L286) mechanism to load the secret key and avoids using `randAlphaNum` secret-key generation, which breaks workflows with tools like argocd. 

Instead, it makes it possible to either use an existing secret for the sentry `secret-key`, or generate one as a helm hook.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 15.0.0
+version: 15.0.1
 appVersion: 22.6.0
 dependencies:
   - name: memcached
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 15.2.2
+version: 15.2.3
 appVersion: 22.8.0
 dependencies:
   - name: memcached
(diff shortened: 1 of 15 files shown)


begemotik:
I'am surprised this one is abandoned. IMHO this is a must to close the security gap when running Helm, there is another open PR for [similar case](https://github.com/sentry-kubernetes/charts/pull/641) . Would love to see this merged.

TaiSHiNet:
@pulse00 do you intend on following up on this? If not I can take over

pulse00:
> @pulse00 do you intend on following up on this? If not I can take over

@TaiSHiNet i fixed the conflicts - this patch is running in our cluster since over a month now without problems.

TaiSHiNet:
Awesome, thanks a lot! Let's hope we can get it merged

begemotik:
@Mokto drawing your attention to this oper PR.

Mokto:
I like the idea although this is a breaking change (would require upgrading the major version)

begemotik:
Makes sense, altough it's not the reason to block the PR, right?

Mokto:
Yes, we have to to upgrade the version before merging.

pulse00:
> Yes, we have to to upgrade the version before merging.

done.

Mokto:
Thanks! Please upgrade to a major version as it contains a breaking change.

PR #675, opened 2022-07-20
Pro 3009 sentry clickhouse security context fix

tuladhp:
added security context to run as nonroot


Mokto:
I don't think this PR is meant to be for this repository?

You've changed a lot of stuff specific to your use case.

PR #679, opened 2022-07-21
Feat: configurable service account for snuba deployments

nishaero:
Currently there is no option to configure service account for individual snuba apps, they all use the same name 'sentry-snuba'. This makes it difficult when these pods try to access external services in Consul mesh.
By adding configurable service account, this allows customizability for individual needs.


Mokto:

PR #681, opened 2022-07-21
Feat: add service account names for cronjobs

nishaero:
Adding an option to provide service account name to cronjobs allows for more customizability.
It allows users to create service accounts if required. 

The current use case is to allow these jobs to register in consul mesh, so that they can access external resources in the mesh like redis, kafka etc...


Mokto:
Thanks. Maybe just bump to 15.0.2 ? ;)

nishaero:
I just bumped the chart version

PR #683, opened 2022-07-21
feat: specifying deadline for jobs and cronjobs

nishaero:
Specifying a deadline for jobs and cronjobs helps manage a failed job to spawn pods indefinately. This sets a time limit until which the job must execute successfully.

Enabling this keeps the cluster clean even when a job fails.


Mokto (approved):

Mokto:
Thanks!

PR #684, opened 2022-07-21
fix: added conditional pod annotations to jobs

nishaero:
In the current hooks templates, there is no option to use only the job specific annotations when we have two annotations specified in the values file.

By using conditionals, we can set either the job specific or app specific annotation. 
This helps in using the two separate annotations without interference.

### DIFF
--- a/sentry/templates/hooks/sentry-db-check.job.yaml
+++ b/sentry/templates/hooks/sentry-db-check.job.yaml
@@ -23,11 +23,10 @@ spec:
     metadata:
       name: {{ template "sentry.fullname" . }}-db-check
       annotations:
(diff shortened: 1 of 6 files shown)


Mokto:
That's a breaking change and I'm not sure to be ready to create a major release only with this change.

Let's keep it on hold until the next major release.

nishaero:
> That's a breaking change and I'm not sure to be ready to create a major release only with this change.
> 
> Let's keep it on hold until the next major release.

Alright lets wait !

nishaero:
@Mokto Can you also please explain why there is two separate pod annotations for this job?

Mokto:

PR #687, opened 2022-07-22
Feat/add serviceaccount names for jobs

nishaero:
Adding an option to provide service account name to hook jobs allows for more customizability.
This option is completely missing in hook jobs

The current use case is to allow these jobs to register in consul mesh, so that they can access external resources in the mesh like redis, kafka etc...


Mokto:

PR #689, opened 2022-07-24
Fix Clickhouse servicemonitor

leo108:
1. Add missing `namespaceSelector` and endpoint `path`
2. Only enable replica servicemonitor when replica is enabled


Mokto (approved):

leo108:
@Mokto Chart version bumpped

Mokto:
Thanks!

PR #691, opened 2022-07-28
Add instructions for cluster of external Kafka instances

aqeelahmad:
This adds an example of how one can configure a cluster of external Kafka instances.


Mokto (approved):

Mokto:
Thanks.

PR #695, opened 2022-08-01
Always create clickhouse config volumes just like volumeMounts

bumarcell:
The volumes should always be created because the config maps are always there (default or custom)
Currently volumeMounts are always created which works fine also with custom configmaps, but the volumes aren't created because of the mentioned if condition


Mokto:

PR #700, opened 2022-08-07
fix: Apply appropriate batch apiVersion based on kubernetes version

yanntoque:
As stated in the [deprecated API Migration Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/\#cronjob-v125), batch/v1beta1 will no longer be served in v1.25
Migrate manifests and API clients to use the batch/v1 API version, available since v1.21 

This commit allows to set the appropriate batch apiVersion based on sentry.kubeVersion

For example if you use : 
*  a  kubernetes version under ```1.21``` it will use ```batch/v1beta1```
*  a kubernetes version  ```>= 1.21``` of kubernetes it will use ```batch/v1```

Signed-off-by: Yann Toqué <toqueyann@gmail.com>

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 15.0.8
+version: 15.0.9
 appVersion: 22.6.0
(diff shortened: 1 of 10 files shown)


Mokto:
Thanks!

PR #702, opened 2022-08-09
Support custom volume mounts everywhere

Agalin:
Allow custom volume mounts for all deployments, stateful sets and cronjobs.

It makes it possible to run properly in restrictive environments with read-only filesystem (where without volume mounted as `/tmp` even logs cannot be read from many of those pods), allows for root ca addition to any pod and many other customizations/workaround (e.g. I'm using it to overwrite some config which contains plaintext credentials to instead take them from env vars - by mounting alternative config files).

Tried to keep indentation consistent where possible.

Also changed one default value -  `snuba.command` from `{}` to `[]` - it's a list so using `{}` generates warnings from helm or skaffold (not sure which one).

### DIFF
--- a/clickhouse/Chart.yaml
+++ b/clickhouse/Chart.yaml
@@ -11,6 +11,6 @@ keywords:
 name: clickhouse
 sources:
 - https://github.com/sentry-kubernetes/charts
-version: 3.1.3
+version: 3.2.0
 maintainers:
   - name: sentry-kubernetes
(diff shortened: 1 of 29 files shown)


Agalin:
@Mokto I know this one is large but any chance for a review?

Agalin:
Found an indent issue in clickhouse, fix for it is also included.

Mokto (approved):

PR #705, opened 2022-08-10
Configurable Security Context for CronJobs

Agalin:
Configurable Security Context is missing from both Snuba cleanup jobs
(cleanup transactions and cleanup errors). It makes it impossible to run
those jobs with `runAsNonRoot` - k8s cannot verify id of a named user.

PR #709, opened 2022-08-12
Configurable resources and env for relay init container

Agalin:
Relay init container is missing option to configure resources. It makes it unusable with strict policies where resources are obligatory.

Additionally added option to configure env vars - don't know relay enough to say how useful it is in this context (config generation) but relay itself supports env variables.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 15.2.1
+version: 15.2.2
 appVersion: 22.8.0
 dependencies:
   - name: memcached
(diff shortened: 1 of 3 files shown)


Mokto (approved):

Agalin:
@Mokto rebased both this and #705. Is that enough to satisfy linter or should I bump version? If so should both PRs bump it separately?

Mokto:
You have to bump the version.

Agalin:
Done. If I understand correctly after this one is merged and #705 rebased it will pass the CI as well.

Agalin:
Ok. Now I'm completely lost. 15.2.0 is not yet released so why is this check failing? Till now it didn't look like each single PR bumps version even without releases.

Mokto:
15.2.0 has been released in the helm chart repository.

Agalin:
I see that 15.2.1 has been built as well. Bumped version to 15.2.2.

Are new charts released directly after merging?

Agalin:
@Mokto any chance for merge while it's still without conflicts? 😄

Mokto:
Thanks!

PR #722, opened 2022-08-30
Add noStrictOffsetReset to consumers

JHeilCoveo:
We frequently have to reset Kafka's log offset in order for Snuba to consume again.
Add a parameter to commands that handle `--no-strict-offset-reset` option.


Mokto:
Really nice, thanks!

PR #746, opened 2022-10-05
Fix sentry features for 22.9.0

andrsp:
in case of deprecation `eventsv2/` endpoint in favor of `events/`, this fixes dashboards and performance frontend parts, that use `eventsv2/` by default


andrsp:
without this fix default dashboards will look like this, same for performance - 404 to eventsv2 endpoint
image

Mokto:
Thanks.

PR #754, opened 2022-10-10
add: support rabbitmq vhost

icewisb:


sentry/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 16.0.2
+version: 16.0.3
```
Mokto: It's a new feature so it should be 16.1.0

PR #757, opened 2022-10-12
Allow for v0 API endpoints

xvilo:
Support the V0 api endpoints.
In the newest release of this chart, sentry 22.9.0. This does API calls to the uri: `example.com/api/0/organizations/[org-name]/eventsv2/?data=here`

With the current set-up this will always result in a 404.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 16.0.2
+version: 16.0.3
 appVersion: 22.9.0
 dependencies:
   - name: memcached
--- a/sentry/templates/ingress.yaml
+++ b/sentry/templates/ingress.yaml
@@ -139,9 +139,9 @@ spec:
               servicePort: {{ template "relay.port" . }}
               {{- end }}
         {{- if eq (default "nginx" .Values.ingress.regexPathStyle) "traefik" }}
-          - path: {{ default "/" .Values.ingress.path }}api/{[1-9][0-9]*}/{(.*)}
+          - path: {{ default "/" .Values.ingress.path }}api/{[0-9][0-9]*}/{(.*)}
         {{- else }}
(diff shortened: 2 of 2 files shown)


Mokto:
Thanks.

xvilo:
@Mokto I was wondering, I suppose this is a fix for 16.x.x version (which we can't use due to other deployment issues). What's the best way to add this change to 15.3.x release? Ideally we'd like to see this released asap as 15.3.1 :)

Mokto:
Not really possible with this branching logic.

Also to be honest I find I already don't have enough time to maintain the project, so maintaining such a system would be too cumbersome.

xvilo:
Would it be possible to create a v15 branch. Then I can create a PR against that. This should allow us to easily tag a new v15 version. And possibly remove the branch thereafter?

reneluria:
This breaks some endpoints like example.com/api/0/organizations/[org-name]/members/[id]/ and gives a 
CSRF Failed: Referer checking failed

It was supposed to be this way at some point
like this:
https://github.com/getsentry/sentry/pull/18433/files#diff-3194b88002aeabc2a7b45a1640d0869bR50

xvilo:
@reneluria i doubt that this causes this, which page gives that? Can you provide more info?

reneluria:
> @reneluria i doubt that this causes this, which page gives that? Can you provide more info?

updating any member access level for exemple, from member to admin:

it's a PUT request on /api/0/organizations/{your_org}/members/{member_id}/

with the ingress matching [0-9], you get "CSRF Failed: Referer checking failed - ... does not match any trusted origins."
with the ingress matching [1-9] like it was before everything is fine

since this update users have been reporting odd issues, reverting this resolved most of them

xvilo:
Interesting, then we should indeed revert this change and see if we can do it differently. I will create a new PR later today

xvilo:
Sorry it took so long, but I've created a revert PR here: https://github.com/sentry-kubernetes/charts/pull/781 

I see it bleeded into the v16 and v17 releases, and will be fixed in the newest v17 release (if merged soon)

PR #766, opened 2022-10-31
split post-process-forwarder

ntcong:
Since https://github.com/getsentry/self-hosted/commit/4a02090030fdaaacbac521784919185afc47d88c, it's required to split post-process-forwarder into errors and transactions entity. This fixed post-process-forwarder deployment crash in the latest sentry (22.10.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,16 @@ Big thanks to the maintainers of the [deprecated chart](https://github.com/helm/
 
 For now the full list of values is not documented but you can get inspired by the values.yaml specific to each directory.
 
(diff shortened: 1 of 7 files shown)


Mokto:
Thanks for your PR. 

1. We should upgrade the major version as it's a breaking change
2. We could also update Sentry to 22.10 at the same time?

ntcong:
> Thanks for your PR.
> 
> 1. We should upgrade the major version as it's a breaking change
> 2. We could also update Sentry to 22.10 at the same time?

I have bumped chart and app version and also added a 17.x upgrade note.

Mokto:
Amazing, thanks!

PR #768, opened 2022-10-31
Custom relay args

Agalin:
Allow custom args in both relay and its init container to enable changing config file used.

Together with #702 it enables full overwrite of the configuration. Why would you want to do that?

* Current one is stored as a configmap instead of a secret while there are redis (and possibly kafka?) credentials included.
* There are at least two formats of the redis configuration block, one for single instance (string) and another (map) for Redis Cluster. Without native support for Redis Cluster throughout the chart it would be quite cumbersome to support it in the template - unless you have a good idea how that could be done cleanly?
* There may be possible other options in existing config blocks that user might want to overwrite / extend. I've tried to verify how does Relay handle repeated keys and to my understanding latest one overwrites the older ones but that's not an intuitive behaviour.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 17.2.0
+version: 17.3.0
 appVersion: 22.10.0
 dependencies:
   - name: memcached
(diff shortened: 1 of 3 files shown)


Mokto (approved):

Mokto:
Thanks for this! Please update the minor version of the chart again (17.2.0).

Agalin:
Those versions go so fast! 😅 Updated.

Mokto:
A lot of people are contributing to this project 🥇

PR #771, opened 2022-11-01
Update configmap-sentry.yaml

Robing:
Add comfig params for s3


Robing:
Add addressing_style and location settings for s3 configuration !

Mokto (approved):

PR #798, opened 2022-12-12
Update sentry to 22.11.0

z0rc:
Ref:
- https://github.com/getsentry/snuba/commit/a77edeac50eaaaac03f1fa69e259d7461eea6ca0
- https://github.com/getsentry/snuba/commit/0fc344533fd358b0a8e9687edf4aa1a9ded2ba46
- https://github.com/getsentry/snuba/commit/543cb78489aa4f98100ddcd98387c5adcfeeadfe

I didn't notice any changes in getsentry/self-hosted between 22.10.0 and 22.11.0, that should be ported here. Also there is https://github.com/getsentry/self-hosted/pull/1817, but it doesn't affect this chart, as we don't set `--commit-log-topic` in `snuba-transactions-consumer` deployment.

Looking volunteers for testing.


pavels:
Hi
i decided to upgrade our system to 22.11 thinking this will be small and simple update and learned the hard way it is not

the whole story is here https://github.com/getsentry/self-hosted/issues/1835#issuecomment-1349928636

TLDR; Transaction stats page stops working because bunch of undocumented stuff is missing - i already have needed modifications for the helm chart so if you like i can provide those

z0rc:
@pavels thanks! I'm okay either way. May I suggest for you to create a PR with all discovered changes? I believe it would be way bigger than current PR, so it makes sense to collaborate around it.

pavels:
Done - see https://github.com/sentry-kubernetes/charts/pull/800

PR #800, opened 2022-12-13
Update sentry to 22.11.0 + undocumented services

pavels:
Here is the pull request fro update to 22.11

it contains finding described here https://github.com/sentry-kubernetes/charts/pull/798#issuecomment-1349940579 and here https://github.com/getsentry/self-hosted/issues/1835#issuecomment-1349928636

One important thing - sentry/templates/deployment-sentry-ingest-metrics-consumer-rh.yaml is redundant now - it does nothing as it produces data, that is not used now - as i understand it, this data will replace release health calculation in the future, but as for now, release health page still reads the old data. I included it anyway as it will most likely be needed in future versions anyway, but it may be removed now as well.


sentry/templates/configmap-relay.yaml
```diff
@@ -37,5 +37,8 @@ data:
       {{- else }}
       redis: "redis://{{ $redisHost }}:{{ $redisPort }}"
       {{- end }}
+      topics:
+        metrics_transactions: ingest-performance-metrics
+        metrics_sessions: ingest-metrics
```
z0rc: trailing spaces

sentry/templates/configmap-relay.yaml
```diff
@@ -37,5 +37,8 @@ data:
       {{- else }}
       redis: "redis://{{ $redisHost }}:{{ $redisPort }}"
       {{- end }}
+      topics:
```
z0rc, line 40: I'm curios whether those topics expected to be handled by kafka's `auto.create.topics.enable` property or actually created by sentry migration/init job?

sentry/templates/configmap-relay.yaml
pavels: ficed

sentry/templates/configmap-relay.yaml
pavels: Interesting question
It appears that only place that creates topics in advance is Snuba - uses this list here https://github.com/getsentry/snuba/blob/22.11.0/snuba/utils/streams/topics.py#L6

This doesn't go through snuba. It is relay -> sentry. It is the same as ingest-consumer (for example `ingest-transactions` topic) - these are created using `auto.create.topics.enable` (on by default because of https://github.com/getsentry/sentry/blob/22.11.0/src/sentry/conf/server.py#L2578)

z0rc (approved):

Mokto:
Thanks, I'll try it now.

z0rc:
According to https://github.com/getsentry/self-hosted/issues/1860#issuecomment-1351318749 we should probably bump symbolicator to 0.6.0 together with sentry 22.11.0.

Mokto (approved):

Mokto:
Thanks!

PR #803, opened 2022-12-20
update Sentry 22.12.0

farodin91:


z0rc:
https://github.com/getsentry/snuba/pull/3468/files#diff-d946011d9919f15a42d92bb5f8fc560631bcad5cc10a20894579d493c7f828daR27 probably should be included in https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/configmap-snuba.yaml#L30

sentry/values.yaml
```diff
@@ -750,6 +750,7 @@ nginx:
 
 ingress:
   enabled: false
+  enabledForwardedHost: false
```
z0rc, line 755: This requires comment what it does, and in what cases it should be used.

farodin91:
all fixes should be done

PR #818, opened 2023-01-16
Add support to specify credentials of clickhouse users with existing secrets

Kasape:
Currently, the usernames and passwords of clickhouse users are passed from values to a configmap. It is not possible to define these values using an existing secret. Password can be defined using SHA256 hash, but it can be hard to maintain. A better option is to have one existing secret with an unhashed password that will be passed to both sentry and clickhouse helm charts. Clickhouse helm chart will load this secret using `lookup` function (as commonly used by bitnami helm charts or recommended by [this guide](https://codersociety.com/blog/articles/helm-best-practices#11-use-the-lookup-function-to-avoid-secret-regeneration)), hash it and insert it to a configmap. That is implemented in this PR. Backward compatibility is preserved.

PR editing sentry chart to also load clickhouse credentials will be submitted separately.

### DIFF
--- a/clickhouse/Chart.yaml
+++ b/clickhouse/Chart.yaml
@@ -11,6 +11,6 @@ keywords:
 name: clickhouse
 sources:
 - https://github.com/sentry-kubernetes/charts
-version: 3.2.1
+version: 3.2.2
 maintainers:
   - name: sentry-kubernetes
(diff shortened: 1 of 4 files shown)


Kasape:
This issue was automatically closed because in the end, we replaced clickhouse from this repository with Bitnami's clickhouse and we are using `externalClickhouse.existingSecret` to pass a password to Snuba

PR #832, opened 2023-02-16
recent changes in snuba break clickhouse 20.8

ntcong:
Since, https://github.com/getsentry/snuba/pull/3604, which is live in sentry 23.1.1 and 23.2, sentry won't process event because of clickhouse error: `QueryExecutionError: DB::Exception: Illegal column for DataTypeNullable.`

Upgrading clickhouse to 20.12 seems to fix the problem, it's a minor version bump so I'm not sure if there's any side-effect. It's running fine on my end though


Mokto:
Thanks, could you patch version please?

ntcong:
> Thanks, could you patch version please?

I increased the version to 18 since a minor clickhouse version bump may or may not cause problem to existing installation

This also requires https://github.com/sentry-kubernetes/charts/pull/833 to fully work though

z0rc:
This must be merged. Sentry from current chart version is half broken without this fix.

PR #840, opened 2023-03-08
Add support for injecting Slack integration secrets via K8s secrets, and fix paths to Bitnami charts

### LINKED ISSUE #829: Add ability to reference existing secret for slack Secrets
Hi there,

The issue had been raised before in https://github.com/sentry-kubernetes/charts/issues/338 with the suggestion to use external secrets. The slack `clientSecret` and `signingSecret` are stored in a configmap in cleartext.

Is it possible to reference existing secrets for it like it has been done for email password in https://github.com/sentry-kubernetes/charts/pull/770/files?

P.S. this is not only true for slack, but also `google` and `github`: https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L713-L734

spectrogram:
- Adds the ability to specify an existing K8s secret containing the parameters for the Slack integration, adds them as environment variables and configures Sentry to use them. Disclaimer: I have not tested this on my installation but the templates turn out alright. This should help resolve https://github.com/sentry-kubernetes/charts/issues/829
- Also, changed the repo URLs for some Bitnami charts that are more than 6 months old. See this issue: https://github.com/bitnami/charts/issues/10545

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,7 +29,7 @@ jobs:
           ct lint \
             --target-branch=develop \
             --chart-dirs=. \
-            --chart-repos=bitnami=https://charts.bitnami.com/bitnami,sentry-kubernetes=https://sentry-kubernetes.github.io/charts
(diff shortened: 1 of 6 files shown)


Mokto (approved):

Mokto:
Really nice, thanks!

PR #828, opened 2023-02-10
allow concurrent cleanup

farodin91:

PR #848, opened 2023-03-25
Implement concurrency

xvilo:
We always have a _very_ big nodestore_node table, and our clean-up cron takes a while. I discovered there is a concurrency option which removes the old NodeStore values a bit faster. 

When not providing the `--concurrency` option, the default used is `1`, so added this is the default in the values. One can override if they need more concurrency


Mokto:
Done in #828 already.

PR #849, opened 2023-03-27
Upgrade Sentry 23.3.1

farodin91:


farodin91:

sentry/templates/deployment-sentry-ingest-replay-recordings.yaml
```diff
@@ -0,0 +1,135 @@
...
+{{- if .Values.sentry.ingestReplayRecordings.volumeMounts }}
+{{ toYaml .Values.sentry.ingestReplayRecordings.volumeMounts | indent 8 }}
+{{- end }}
+        resources:
+{{ toYaml .Values.sentry.ingestReplayRecordings.resources | indent 12 }}
+{{- if .Values.sentry.ingestReplayRecordings.sidecars }}
+{{ toYaml .Values.sentry.ingestReplayRecordings.sidecars | indent 6 }}
+{{- end }}
+      {{- if .Values.serviceAccount.enabled }}
+      serviceAccountName: {{ .Values.serviceAccount.name }}-ingest-replay-recordings
```
z0rc, line 107: Missing template for this ServiceAccount.

farodin91: done

farodin91:
@z0rc  Would you like to review it again?

z0rc (commented):
I am not a maintainer, and don't have ability to merge.

Also it's probably worth to mention somewhere that cronjobs were removed and why this happened. Also personally at this case I'd bump major chart version, but this is up to chart maintainer to decide.

pavels:
There is probably more things happening around max batch sizes - as they are moving more towards arroyo, they are dropping the batch size arguments almost everywhere

Billing consumer for example doesn't have it as well now, and probably more - according to https://github.com/getsentry/sentry/pull/45171#issuecomment-1448775163 - `in nearly all Sentry consumers today we just commit once per second - imo this should be all consumers eventually`

It would be also probably nice to mention this somewhere? Some changelog? We for example rely heavily on the maxBatchSize / commitBatchSize settings for performance and i was a bit scared when i saw these removed - until i went through bunch of sentry PR to find out that they are actually removing these completely.

stevenyongzion:
When upgrading changes, we encounter the following errors:

`Error: UPGRADE FAILED: template: sentry/templates/deployment-snuba-subscription-consumer-sessions.yaml:26:22: executing "sentry/templates/deployment-snuba-subscription-consumer-sessions.yaml" at <.Values.snuba.subscriptionConsumerSessions.replicas>: nil pointer evaluating interface {}.replicas`

In the value.yaml we think the following should be in the Snuba section as well as Sentry section:

```
  subscriptionConsumerSessions:
    replicas: 1
    env: []
    resources: {}
    affinity: {}
    nodeSelector: {}
    securityContext: {}
    # tolerations: []
    # podLabels: []
    # commitBatchSize: 1
    sidecars: []
    volumes: []
    # noStrictOffsetReset: false
    # volumeMounts: []  
```

`autoOffsetReset: "earliest"` should be added as well else the pod will throw the following errors:

`Error: Invalid value for '--auto-offset-reset': '' is not one of 'error', 'earliest', 'latest'.`

Additionally, the pod `sentry-ingest-replay-recordings` will throw these errors:

```
arroyo.errors.ConsumerError: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: ingest-replay-recordings: Broker: Unknown topic or partition"}
12:20:10 [ERROR] arroyo.processing.processor: Caught exception, shutting down...
```

farodin91:
@stevenyongzion  i fixed you issue

Mokto:
Looks good, thanks! I'll merge as soon as the linting issues are fixed.

farodin91:
@Mokto  Issues fixed

Mokto:
Thanks a lot!

z0rc:
After updating to this version, individual issue view didn't work in sentry. Cause and fix is in https://github.com/sentry-kubernetes/charts/pull/832

farodin91:
@z0rc i'm using an external managed clickhouse.

alexandresoro:
Hi,

I upgraded fom 17.11.3 to 18, and have a similar log as @stevenyongzion mentioned on: `sentry-ingest-replay-recordings`.

```
arroyo.errors.ConsumerError: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: ingest-replay-recordings: Broker: Unknown topic or partition"}
```

Is there any manual action to perform during the upgrade?

adonskoy:
> Hi,
> 
> I upgraded fom 17.11.3 to 18, and have a similar log as @stevenyongzion mentioned on: `sentry-ingest-replay-recordings`.
> 
> ```
> arroyo.errors.ConsumerError: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: ingest-replay-recordings: Broker: Unknown topic or partition"}
> ```
> 
> Is there any manual action to perform during the upgrade?

Faced the same problem. To solve, it is enough to create a topic by yourself.

Looks like a problem in snuba migrations in v23.3.1

pehlert:
fyi, I created the topic by opening a shell on the kafka-0 pod and running:

```
/opt/bitnami/kafka/bin/kafka-topics.sh --create --topic ingest-replay-recordings --bootstrap-server localhost:9092
```

I can't guarantee that it's enough to make it work, but it at least fixes the crashes.

PR #876, opened 2023-05-09
Use pod template for snuba

farodin91:
I would like to reduce the overhead to write deployments step by step.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -9,6 +9,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: "0"
+      - uses: d3adb5/helm-unittest-action@v2
+        with:
+          helm-version: v3.8.0
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          charts: sentry
       - name: Set up chart-testing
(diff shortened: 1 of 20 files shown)


farodin91:
@Mokto  Would you like to review this PR?

Mokto:
I'm not super confident on making that change. I feel like we're losing flexibility and more importantly, it's more prone to errors without a proper test suite.

farodin91:
I will try to built a proper test suite for it. 

I tried to reduce the error prone behavior by moving everything into a single helper function.

Mokto:
If you do that, that would be perfect 👍

farodin91:
@Mokto  I added a basic idea for unittest. What do you think? 

My current concern only snuba-api tested. I would like to find a way to test all snuba without duplication.

farodin91:
I could do a snapshot test this would require user manually review rendered data.

Mokto:
I'd rather have snapshots so we know exactly what changes when anything changes in the repository.

Thanks!

farodin91:
@Mokto  I added three snapshot tests. Should i add them for all deployments? Or should we split tests and helper changes.

Mokto:

PR #857, opened 2023-04-07
feat: Options to setup a custom CA for sentry-kubernetes

autowp:
Including REQUESTS_CA_BUNDLE and update of `/usr/local/lib/python3.7/site-packages/certifi/cacert.pem` for urllib3

PR #696, opened 2022-08-01
Ability to specify customCA for worker deployment

hardenchant:
Add ability to add customCA to celery worker deployment like for web deployment.
According to this comment:
https://github.com/sentry-kubernetes/charts/issues/371#issuecomment-1193864927

Issue #371, opened 2021-05-03
Running on-prem using self-signed TLS certs


bjschafer:
We're attempting to use this chart with a TLS cert signed by our internal root CA, and seeing the below errors in the `sentry-web` pod logs.

We have populated the `sentry.web.customCA` value, and set a known-good CA chain in the secret listed in `sentry.web.customCA.secretName`. Note that the `item` property in `customCA` causes an error to be thrown [here](https://github.com/sentry-kubernetes/charts/blob/13f4537e4df3d67b4be82234309a580d2c4c9af4/sentry/templates/deployment-sentry-web.yaml#L141) because `item` isn't a property of a `Secret`.

Digging into the stacktrace a little, it seems to be calling the SDK which doesn't seem to respect the CA we're passing in.

We're currently unable to use a publicly trusted cert, so that workaround doesn't apply here.

---

```
10.128.101.0 - - [03/May/2021:21:45:30 +0000] "GET /_health/ HTTP/1.1" 200 203 "-" "kube-probe/1.19"
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/sentry/../urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/usr/local/lib/python3.6/site-packages/sentry/../urllib3/connectionpool.py", line 343, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.6/site-packages/sentry/../urllib3/connectionpool.py", line 839, in _validate_conn
    conn.connect()
  File "/usr/local/lib/python3.6/site-packages/sentry/../urllib3/connection.py", line 344, in connect
    ssl_context=context)
  File "/usr/local/lib/python3.6/site-packages/sentry/../urllib3/util/ssl_.py", line 347, in ssl_wrap_socket
    return context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/local/lib/python3.6/ssl.py", line 407, in wrap_socket
    _context=self, _session=session)
  File "/usr/local/lib/python3.6/ssl.py", line 817, in __init__
    self.do_handshake()
  File "/usr/local/lib/python3.6/ssl.py", line 1077, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/local/lib/python3.6/ssl.py", line 689, in do_handshake
    self._sslobj.do_handshake()
[...]

PR #879, opened 2023-05-11
feat: Options to setup a custom CA for sentry web & worker

autowp:
Including REQUESTS_CA_BUNDLE and update of /usr/local/lib/python3.8/site-packages/certifi/cacert.pem for urllib3

Alos fix issue with customCA:
`Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.volumes[2].secret): unknown field "item" in io.k8s.api.core.v1.SecretVolumeSource`

Related to #696
Same for sentry-kubernetes: #857


christiangonre:
@autowp can you take a look and solve the conflicts from this PR?

I replicated this in my local environment and it works fine, maybe after fixing the conflicts @Mokto can take a look into the MR and merge it. If @autowp cannot solve the conflict, I can create a new MR from develop to make it work.

sentry/templates/deployment-sentry-worker.yaml
```diff
@@ -77,6 +92,10 @@ spec:
         - name: C_FORCE_ROOT
           value: "true"
 {{ include "sentry.env" . | indent 8 }}
+        {{ if .Values.sentry.worker.customCA }}
+        - name: REQUESTS_CA_BUNDLE
+          value: /etc/pki/ca-trust/custom/{{ default "ca.crt" .Values.sentry.worker.customCA.item }}
+        {{ end }}
```
christiangonre:
```suggestion
```
This is not needed because we will package all the CA into a single file, so the REQUEST library will be able to read the new CustomCA from there.

Same for the sentry-web deployment, we should remove this variable from there.

autowp:
@christiangonre I'm sorry for the late reply

Rebased on develop and took into account your remark against REQUESTS_CA_BUNDLE

sentry/values.yaml
```diff
@@ -179,6 +179,9 @@ sentry:
     nodeSelector: {}
     # tolerations: []
     # podLabels: []
+    # customCA:
```
radubontila, line 182: I would add customCA value to sentry level and use it everywhere from there. Usually if you need custom CA for web you will need it for worker as well. Also deployment-sentry-cron needs the same changes, it calls relay api and as far as I discovered deployment-sentry-ingest-metrics-consumer-perf and deployment-sentry-ingest-metrics-consumer-rh have ssl errors if you need customCA, so same changes here as well

patsevanton:
Any updates?

chipzzz:
This is very much needed, the problem exists and it's difficult to overwrite the worker ca path. Even trying with 

https://github.com/sentry-kubernetes/charts/issues/371#issuecomment-1641235675

Mokto:

calebmayeux:
For the next poor sucker that has to set up sentry in a corporate environment with their own CA, this is how I worked around it without having to rebake images or fork helm charts.

First, create a Kubernetes secret called corporate-ca whose contents are a key ca.crt and the value base64 encoded private CA file. Then, pass these variables to your helm chart. The web container kinda does the work for you, but for the cron and worker pods, you have to set environment variables to get python to read your CA and mount your secret as a file inside the pods.

```
sentry:
  web:
    customCA:
      secretName: corporate-ca
  worker:
    env:
      - name: REQUESTS_CA_BUNDLE
        value: /etc/ssl/certs/ca.crt
      - name: SSL_CERT_FILE
        value: /etc/ssl/certs/ca.crt
    volumes:
      - name: corporate-ca-volume
        secret:
          secretName: corporate-ca
    volumeMounts:
      - name: corporate-ca-volume
        mountPath: /etc/ssl/certs/ca.crt
        subPath: ca.crt
  cron:
    env:
      - name: REQUESTS_CA_BUNDLE
        value: /etc/ssl/certs/ca.crt
      - name: SSL_CERT_FILE
        value: /etc/ssl/certs/ca.crt
    volumes:
      - name: corporate-ca-volume
        secret:
          secretName: corporate-ca
    volumeMounts:
      - name: corporate-ca-volume
        mountPath: /etc/ssl/certs/ca.crt
        subPath: ca.crt
 ```
If somebody wants to make this better, they could basically put the logic for the worker and cron in the helm chart and just have a sentry customCA field that when set does all the heavy lifting.

patsevanton:
@calebmayeux please write how do you see how you can shorten values.yaml? However, I do not know if this will happen.

calebmayeux:
> @calebmayeux please write how do you see how you can shorten values.yaml? However, I do not know if this will happen.

SSL_CERT_FILE may be redundant, no time to test it. Other than that, probably would require fixing the helm chart.

Another thing to note when doing this, the worker tries to pull a model file from the internet, and repointing the trusted CA cert means it will fail to trust the site's certificate when it tries to pull that. This is moot if your instance is not connected to the Internet though.

To get around that issue I made my own Dockerfile based off the sentry image and baked the model into it, so no internet or pull needed.

Based on instructions from here: https://stackoverflow.com/questions/76106366/how-to-use-tiktoken-in-offline-mode-computer

FROM repo-redacted/getsentry/sentry:24.8.0
RUN mkdir /opt/tiktoken
ADD https://openaipublic.blob.core.windows.net/encodings/cl100k_base.tiktoken /opt/tiktoken/9b5ad71b2ce5302211f9c61530b329a4922fc6a4
ENV TIKTOKEN_CACHE_DIR=/opt/tiktoken

Issue #743, opened 2022-10-04
Getting Too many parts (300). Merges are processing significantly slower than inserts from clickhouse


brianchan661:
My sentry is handling around 700k error per hour, and was running well for 5 months,
while suddenly I start getting below errors from click house.
Tried to delete most of the old data in click house pvc, but doesn't help.
Do some one have idea how should i solve it?

> 2022-10-04 13:34:42,299 librdkafka log level: 6
2022-10-04 13:34:45,094 New partitions assigned: {Partition(topic=Topic(name='events'), index=0): 27744513, Partition(topic=Topic(name='events'), index=1): 33082072, Partition(topic=Topic(name='events'), index=2): 46448020, Partition(topic=Topic(name='events'), index=3): 11878832, Partition(topic=Topic(name='events'), index=4): 39330301, Partition(topic=Topic(name='events'), index=5): 29221828, Partition(topic=Topic(name='events'), index=6): 17687435, Partition(topic=Topic(name='events'), index=7): 11550491, Partition(topic=Topic(name='events'), index=8): 27825289, Partition(topic=Topic(name='events'), index=9): 35082427, Partition(topic=Topic(name='events'), index=10): 11887986, Partition(topic=Topic(name='events'), index=11): 30928059, Partition(topic=Topic(name='events'), index=12): 16353313, Partition(topic=Topic(name='events'), index=13): 16767766, Partition(topic=Topic(name='events'), index=14): 12629837, Partition(topic=Topic(name='events'), index=15): 9684161, Partition(topic=Topic(name='events'), index=16): 25702358, Partition(topic=Topic(name='events'), index=17): 42009924, Partition(topic=Topic(name='events'), index=18): 31760085, Partition(topic=Topic(name='events'), index=19): 16385015}
2022-10-04 13:34:46,020 Caught {
  "__type__": "SerializableException",
  "__name__": "ClickhouseWriterError",
  "__message__": "Received from sentry-prod-clickhouse-2.sentry-prod-clickhouse-headless.sentry--prod.svc.cluster.local:9000. DB::Exception: Too many parts (301). Merges are processing significantly slower than inserts.. (version 20.8.19.4 (official build))",
  "__should_report__": true,
  "__extra_data__": {
    "code": 252,
    "row": null
  }
}, shutting down...
2022-10-04 13:34:46,020 Terminating <snuba.consumers.consumer.ProcessedMessageBatchWriter object at 0x7f3b0dcb8f40>...
[...]

PR #884, opened 2023-05-16
feat: Allow to set specified merge_tree settings in config.xml

randrusiak:
A few weeks ago I faced a problem with [clickhouse](https://github.com/sentry-kubernetes/charts/issues/743).
The solution was to add [merge_tree setting](https://github.com/sentry-kubernetes/charts/issues/743#issuecomment-1312918533) to Clickhouse `config.xml`.

Unfortunately, it is not possible to set above settings via chart values so I decided to add handling these specific params.


Mokto (approved):

Mokto:
Thanks a lot!

PR #888, opened 2023-05-23
feat: updated kafka helm chart and supports kafka without zookeeper

Mokto:


adonskoy:
@Mokto, sentry probably still uses version 2.5.x, and you want to upgrade it to 3.2, so it's worth running e2e tests before merging

Mokto:
Thanks for letting me know! For now I'm running it without any issues, I'll probably wait for at least a week before anything else.

Would you care to review my PR ? @adonskoy

sentry/Chart.yaml
```diff
@@ -14,9 +14,10 @@ dependencies:
     version: 16.12.1
     condition: redis.enabled
   - name: kafka
-    repository: https://raw.githubusercontent.com/bitnami/charts/pre-2022/bitnami
-    version: 16.3.2
+    repository: oci://registry-1.docker.io/bitnamicharts
+    version: 22.1.3
     condition: kafka.enabled
+    alias: kafka2
```
adonskoy: Do not change the dependency name. I noticed that in the Kafka chart from version 16 to version 21 "breaking" changes only in the zookeeper subchart in the auth section. Using an alias only complicates the update, since all kafka resources will be recreated

It is enough to indicate in the readme what you should pay attention to during upgrade, and what you need to do in order to stay on the zookeeper (`kafka.kraft.enabled` must be `false` and `kafka.zookeeper.enabled` must be `true`)

sentry/Chart.yaml
adonskoy, line 1: Maybe we can update other chart dependencies in this MR, if possible? In order not to stretch it over several major releases, but to refresh the dependencies in one version

adonskoy (changes requested):

sentry/Chart.yaml
adonskoy: redis, memcached and nginx can be updated without problems. But for rabbitmq, need to make a couple of changes in the templates

sentry/Chart.yaml
Mokto: I'll try the upgrade path to make sure it's possible.

sentry/Chart.yaml
Mokto: I honestly don't have time to do that here.

sentry/Chart.yaml
Mokto: 👍

sentry/Chart.yaml
adonskoy: ok, I will update the remaining dependencies in the near future

adonskoy:
@Mokto, you have published your chart from this PR to an index file tagged 19.0.0 and 19.1.0. I believe this was done by accident and needs to be rolled back as soon as possible.

Mokto:
I don't think so ?

19.1.0 is this one: https://github.com/sentry-kubernetes/charts/commit/6bdcea8690d207b29e610f28363841cf0b0912c9

adonskoy:
> I don't think so ?
> 
> 19.1.0 is this one: [6bdcea8](https://github.com/sentry-kubernetes/charts/commit/6bdcea8690d207b29e610f28363841cf0b0912c9)

My apologies, only in version 19.0.0

In the chart index file in version 19.0.0 has Kafka dependency with version 22.1.3 and new OCI repo, but in the commit [`47bc4c1`](https://github.com/sentry-kubernetes/charts/commit/47bc4c1224b68c5fda4f486921356ca64ff0e75a) it's v16.3.2. The dates don't match either. Commit was on May 18, and in the index file the version was published with the date May 23.
image

Mokto:
Just checked, the upgrade doesn't work unfortunately.

ERROR ==> KRaft requires KAFKA_CFG_NODE_ID to be set for the quorum controller

I suspect something is saved in the persistent volume.

adonskoy:
@Mokto,

The problem seems to be in envs of Kafka image. See https://github.com/bitnami/containers/pull/34152

Mokto:
Not sure it's this case tbh.

It works fine on a fresh new install.

adonskoy:
The issue for which this pr was created contains the same error that you received. But that seems to be the least of the problems with the upgrade: Kafka's documentation says that automatic migration from zookeeper to Kraft is in early beta. 

When kafka releases a version with a stable migration mechanism, then it will be possible to enable kraft by default

Maybe just update the major version of kafka and keep zookeeper enabled? Or switch to kraft and mention in the readme the need of removing PVCs during upgrade? The first option is better in my opinion

Mokto:
What don't you actually like with my approach ? I could call it something like `kafka-kraft` so it makes more sense, maybe?

The only downside is potential data loss, but as long as it's run during a quiet period, should be fine.

adonskoy:
The alias idea minimizes the number of manual actions during the update, but this will involves data loss (or backup and restore of data, which complicates the update), and this can be a big problem for some users

One of my sentry installations processes up to a million events a day. I would not really like to lose messages that consumers do not have time to read from topics during the update.

I think it's more correct to use a scenario where the kafka version is updated, but the zookeeper remains. If someone wants to switch to kraft, he can set the values for himself and remove the PVCs. When automatic migration to kraft becomes stable, then I think bitnami will turn on kraft by default in their chart themselves and we will only need to update the subchart version.

Mokto:
Ok you convinced me.

sentry/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 19.3.0
```
adonskoy: Probably this change should be a new major release? The major version for kafka and zookeeper changes between 16 and 22 version of Kafka chart

adonskoy:
@Mokto, you missed to specify in values that kraft is disabled by default

Mokto:
Sorry for the delay, I was off.

Fixed it on 20.1.1

PR #890, opened 2023-05-24
Ingress Path Matching In Traefik

Peytob:
Need to change path to another view, [as described in Github](https://github.com/traefik/traefik/issues/5606)

Example:
path: /api/{id:[1-9][0-9]*}/{rest:(.*)}

regex in Traefik works only when we named it as a variable.

### DIFF
--- a/sentry/templates/ingress.yaml
+++ b/sentry/templates/ingress.yaml
@@ -139,7 +139,7 @@ spec:
               servicePort: {{ template "relay.port" . }}
               {{- end }}
         {{- if eq (default "nginx" .Values.ingress.regexPathStyle) "traefik" }}
-          - path: {{ default "/" .Values.ingress.path }}api/{[1-9][0-9]*}/{(.*)}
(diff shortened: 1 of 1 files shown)


Mokto:
I'm not using Traefik, so it's a bit difficult to review for me. Can anyone else review that?

z0rc:
~This is specific to traefik v1 only, which isn't supported anymore, https://doc.traefik.io/traefik/deprecation/releases/#versions~

~Current traefik v2 doesn't have this requirement. Therefor this PR isn't needed.~

Seems it is needed after all, see https://doc.traefik.io/traefik/routing/routers/#rule, `Regexp Syntax` section.

But with upcoming traefik v3 it won't be needed anymore, see https://github.com/traefik/traefik/issues/9520, and must be reverted.

prom3theu5:
This is definitely needed atm when using traefik v2 (like bundled with k3s)
I just stumbled upon this because I was getting CSRF errors on the initial run setup page after login.

Changing the regex like in this PR solved it

Mokto:
As Traefik 3 has been released, I'm closing this. Feel free to reopen if needed.

PR #895, opened 2023-05-25
Dependency upgrade

adonskoy:
Updated charts: memcached, redis, rabbitmq and nginx. No breaking changes that affects chart templates

Also updated sentry to 23.5.1


adonskoy:
@Mokto, would you like to review? There is also a patch update for sentry itself

Mokto (approved):

PR #904, opened 2023-06-07
add container security context

pschwager:
We are currently trying to deploy this chart to our very restrictive environment. This environment, requires to specify the `allowPrivilegeEscalation` which is only valid on container level security context. Currently the sentry and snuba charts only allow to specify the security context of the pod. With this MR I adapted the charts to specify the security context for the pod and container, like its also done in the bitnami charts.

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 19.3.0
+version: 19.4.0
 appVersion: 23.6.1
 dependencies:
   - name: memcached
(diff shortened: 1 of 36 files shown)


pschwager:
@Mokto May I ask you to take a look at it? Let me know if there's something you would like to be adapted :)

pschwager:
Any chance to get this merged?

Mokto:
This PR is :

1. Breaking all our users charts because changing names in values.yaml
2. Forcing some specific values set to all the people using that chart

If you want to add the ability to add `containerSecurityContext`, you can make that change but please don't introduce breaking changes.

And don't force values in values.yaml.

pschwager:
Thanks for your reply @Mokto 
So I throw away the renaming of the `securityContext` to `podSecurityContext` and only add a new value `containerSecurityContext`. The values.yaml only adds new empty `containerSecurityContext`, where a `securityContext` already exists.

For the hooks I added a fallback to either use specific `.Values.hooks.dbInit.securityContext` or if that is not set the old `.Values.hooks.securityContext`, so it's backward compatible and allows different settings for the snuba and sentry hooks

Mokto:
Looks good thanks!

PR #914, opened 2023-06-20
Extend Redis functionality

frank-m:
- Adds support to set REDIS_PASSWORD as a kubernetes secret.
- Adds support to set externalredis.db too choose the db number
- Adds support to use externalredis.ssl for secure ssl
- Adds support to set externalredis.brokerUrl, this overrides the BROKER_URL environment variable that is used to set the broker with a secret.

The changes are backwards compatible/optional so they won't break existing setups.

### DIFF
--- a/sentry/templates/configmap-relay.yaml
+++ b/sentry/templates/configmap-relay.yaml
@@ -1,6 +1,8 @@
-{{- $redisHost := include "sentry.redis.host" . -}}
-{{- $redisPort := include "sentry.redis.port" . -}}
-{{- $redisPass := include "sentry.redis.password" . -}}
+{{- $redisHost   := include "sentry.redis.host" . -}}
+{{- $redisPort   := include "sentry.redis.port" . -}}
+{{- $redisPass   := include "sentry.redis.password" . -}}
+{{- $redisDb     := include "sentry.redis.db" . -}}
+{{- $redisProto  := ternary "rediss" "redis" (eq (include "sentry.redis.ssl" .) "true")  -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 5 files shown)


frank-m:
Not sure why linting is failing. I have ran it locally and the linting seemed fine..

maciejp-ro:
Is this PR missing anything? What's preventing this change from being merged?

I discovered this PR when preparing to send a similar one (smaller, just TLS support, lack of which is a blocker for me). If the way this PR implements it is not mergeable, please explain what's the problem here and what kind of implementation would be acceptable. Thank you!

sentry/templates/configmap-relay.yaml
```diff
@@ -1,6 +1,8 @@
-{{- $redisHost := include "sentry.redis.host" . -}}
-{{- $redisPort := include "sentry.redis.port" . -}}
-{{- $redisPass := include "sentry.redis.password" . -}}
+{{- $redisHost   := include "sentry.redis.host" . -}}
+{{- $redisPort   := include "sentry.redis.port" . -}}
+{{- $redisPass   := include "sentry.redis.password" . -}}
```
maciejp-ro, line 3: I see that definition of `sentry.redis.password` haven't changed – how will it pick up `externalRedis.existingSecret`?

paco-sisutech:
looking forward to this since we are stuck trying to add an external redis but we would really like to use secrets for the password

sentry/templates/configmap-relay.yaml
frank-m: Unfortunately I could not get redis password in the relay config. But I solved that by setting a redis url. if that env var is set, the password is not needed.

```
  env:
    - name: RELAY_REDIS_URL
      valueFrom:
        secretKeyRef:
          name: sentry-secret
          key: sentry-redis-url
```

felipeng:
looking forward for these options

felipeng:
@frank-m are you planning to work on it?

Mokto:

Issue #867, opened 2023-04-25
Snuba pods are in `Error` and `CrashLoopBackOff` state when `Clickhouse` pod restarted


kseniyashaydurova:
### Problem
Currently there is situation that when any of `sentry-clickhouse` pods is restarted, there appears  crashed snuba related pods

![image](https://user-images.githubusercontent.com/26448672/234196312-a2b78afb-e8bb-45ba-a657-b9a759e252cd.png)

**Can you please explain is it possible to avoid somehow such a behavior?**

### Expected

No pods would fail if any of `Clickhouse` replicas is restarted, because clickhouse is configured as replicaset (as I understand), so snuba should take any other available pod if one becomes not available. But now snuba restarts if any of `clickhouse` pods restarted that doesn't look sensible for HA approach.

### Logs
```
✗  klo sentry-snuba-transactions-consumer-b4d77f6b7-n7sbc --previous
2023-04-25 06:42:37,541 Initializing Snuba CLI...
2023-04-25 06:42:37,544 Initializing Snuba
2023-04-25 06:42:38,939 Snuba initialization took 1.396822193171829s
2023-04-25 06:42:38,943 Consumer Starting
2023-04-25 06:42:38,944 librdkafka log level: 6
2023-04-25 06:42:38,955 New partitions assigned: {Partition(topic=Topic(name='transactions'), index=0): 27728}
2023-04-25 06:42:42,785 Caught exception, shutting down...
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/arroyo/processing/processor.py", line 180, in run
    self._run_once()
  File "/usr/local/lib/python3.8/site-packages/arroyo/processing/processor.py", line 220, in _run_once
    self.__processing_strategy.poll()
  File "/usr/local/lib/python3.8/site-packages/arroyo/processing/strategies/transform.py", line 63, in poll
    self.__next_step.poll()
  File "/usr/local/lib/python3.8/site-packages/arroyo/processing/strategies/collect.py", line 135, in poll
    self.close_and_reset_batch()
  File "/usr/local/lib/python3.8/site-packages/arroyo/processing/strategies/collect.py", line 201, in close_and_reset_batch
    self.future.result(timeout=self.wait_timeout)
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
[...]

PR #915, opened 2023-06-21
use internal_replication in shard

### LINKED ISSUE #820: internal_replication seem to be set in wrong place
Chart version 17.9.0.
According to clickhouse documentation https://clickhouse.com/docs/en/engines/table-engines/special/distributed/#distributed-clusters `internal_replication` is a shard wide parameter not replica specific:
```
        <shard>
            <!-- Optional. Shard weight when writing data. Default: 1. -->
            <weight>1</weight>
            <!-- Optional. Whether to write data to just one of the replicas. Default: false (write data to all replicas). -->
            <internal_replication>false</internal_replication>
            <replica>
                <!-- Optional. Priority of the replica for load balancing (see also load_balancing setting). Default: 1 (less value has more priority). -->
                <priority>1</priority>
                <host>example01-01-1</host>
                <port>9000</port>
            </replica>
            <replica>
...
```
But actual configMap in my setup:
```
    <clickhouse_remote_servers>
        <sentry-clickhouse>
            <shard>
                <replica>
                    <internal_replication>true</internal_replication>
                    <host>sentry-clickhouse-0.sentry-clickhouse-headless.sentry.svc.cluster.local</host>
                    <port>9000</port>
                    <user>default</user>
                    <compression>true</compression>
                </replica>
                <replica>
```
Also this parameter look messy in chart. README states default is `false` but then it redefined in `values.yaml`
```
charts/clickhouse/README.md:| `clickhouse.configmap.remote_servers.internal_replication`                        | If this parameter is set to 'true', the table where data will be written is going to replicate them itself                     | `false`                                               |
charts/clickhouse/values.yaml:      internal_replication: false
values.yaml:        internal_replication: true
```

zetaab:
fixes #820


sentry/values.yaml
```diff
@@ -933,7 +933,7 @@ clickhouse:
     imageVersion: "20.12.8.5"
     configmap:
       remote_servers:
-        internal_replication: true
+        internal_replication: false
```
zetaab, line 1154: because this has been in incorrect place - it means that the default behaviour has been `false` for everyone. That is why this should be false, and also docs says that default value is false.

begemotik:
@zetaab is that a related issue you trying to solve https://github.com/sentry-kubernetes/charts/issues/867#issuecomment-1600658950?

zetaab:
@begemotik well I tried to make clickhouse setup highly available, but the thing is that the clickhouse setup is not installed correctly.

Ok I can try to explain what is wrong with clickhouse helm chart (in this repo):

from https://clickhouse.com/docs/en/engines/table-engines/special/distributed

> Shards refer to the servers that contain different parts of the data (in order to read all the data, you must access all the shards). Replicas are duplicating servers (in order to read all the data, you can access the data on any one of the replicas).

when we check the configuration that helm produces: https://gist.github.com/zetaab/835cd5b58f5a4eb84d8d308ed0947f9d

this means that we have 3 shards with only 1 replica. It means that if we lose even one shard (because shard has only one replica its enough to lose 1/3 of pods), the whole service will be unavailable. 

There is way to define backup replicas for clickhouse https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L939

zetaab:
need to test tomorrow, is there something that can be done to make it work.

Mokto:
Thanks for that! How does the migration path work? Do we need to do anything?

zetaab:
@Mokto there is no need to do any migrations, because this has not been working before. It has used the default behaviour which is false

z0rc:
@zetaab what would be the process for enabling `internal_replication` for existing installation after this MR is merged? Would setting `clickhouse.clickhouse.configmap.remote_servers.internal_replication=true` be sufficient or some kind of data migration is needed?

zetaab:
@z0rc clickhouse does not create table structures to replicas. After those are created somehow, it will work like normal clickhouse replica cluster.

Mokto:
I'm a bit confused about your statement `After those are created somehow`. What would trigger the replication?

zetaab:
that is how clickhouse cluster replication works. It is not enough that you have cluster replica, it should contain similar table structures as well. After that it will start replicating the data inside the cluster

Mokto:
But who takes care of creating those similar structures then?

zetaab:
when creating the setup from scratch by using this helm chart, the snuba migration will create those. However, if this is turned on to existing setup - then the migration needs to be done manually to each replica.

Mokto:
Could you please add some documentation on how to do that?

adonskoy:
It would be nice to automate this process or provide guide, however we already have problem with snuba migrations in the chart after [this PR](https://github.com/sentry-kubernetes/charts/pull/931) was accepted.

Snuba allows us to [apply migrations to a single cluster node](https://github.com/getsentry/snuba/blob/master/MIGRATIONS.md#adding-a-new-node). If it works, then we can add/change snuba migration hook that will apply them to each node separately, and then remove it after 1-2 major versions

adonskoy:
I took a closer look at clickhouse replication and found out the following:
- the current config assumes that the cluster has 3 shards with 1 replica in each. In order for each shard to replicate to each server, we need all N servers to be listed in each shard. Without this, internal replication makes no sense.
- There are currently no easy options for creating replicas for existing tables. Instead of doing this: creating a dump in sql format, deleting clickhouse volumes, applying snuba migrations to a fresh cluster and inserting data from the dump. So the data will be saved, but it requires downtime

zetaab:
anyway I would like to see this merged. Migrating from existing setup to working setup is difficult like said. BUT if someone starts from scratch this is needed. Also we are currently always manually modifying this configmap that we can get replicas working.

begemotik:
I guess more chances to get it merged if some additional documentation around that is added

begemotik:
@zetaab, correct me if I'm wrong. Since you defaulted internal_replication to `false` there won't be broken clusters for end users if they update Sentry helm chart. But when deploying from scratch if values are set as below the HA setup should work out of the box.

```
    clickhouse:
      clickhouse:
        configmap:
          remote_servers:
            internal_replication: true
            replica:
              backup:
                enabled: true
```

If that is right, I think it's quite safe to merge it.

zetaab:
@begemotik right.

The situation is that currently the config is in incorrect place. Which means that all current clusters will use the default value which is `false`. So adding this to correct place with value `false` does not change anything. However, this makes it possible that people can deploy replicas from scratch and the replication will really work.

The only thing that we are lacking here is migration guide to replicas to EXISTING setup. However, that is not so easy like other people mentioned in this PR as well. I have done it, but no idea what is the correct way really.

begemotik:
Yesterday upgraded Sentry installation from 17.8.0 chart version to the most recent and stumbled upon failed Clickhouse migrations. As result I dumped all the data from clickhouse shards with [clickhouse-backup](https://github.com/Altinity/clickhouse-backup/blob/master/Examples.md#how-to-make-backup--restore-sharded-cluster) and restored on the new Clickhouse cluster with "backup replicas" enabled

```
clickhouse:
    clickhouse:
      configmap:
        remote_servers:
          internal_replication: true
          replica:
            backup:
              enabled: true
```

I must note that the cluster is fault tolerant now, but I didn't have to modify the ConfigMap in the way suggested in the MR. If I restart any of 2 replicas of one shard - UI is not failing and works without downtime.

@zetaab could you double check and confirm that ConfigMap adjustments are indeed required? If so, I wonder why it works in my case.

zetaab:
@begemotik I just started vacation, I am back in 1.5 weeks. I cannot test before that easily.

chipzzz:
[begemotik](https://github.com/begemotik) [Mokto](https://github.com/Mokto) [zetaab](https://github.com/zetaab)

How is this even a debate and this PR not merged?! This fixes the incorrect placement of `internal_replication` in the CH config.

Today the helm chart still shows true. So anyone coming into chart is thinking this is set true by default. This could be drastic when trying to fully fix distribued/sharded mode of snuba/clickhouse. Based on the findings mentioned in the official CH docs the placement of this config is entirely incorrect so this has been always false as mentioned by @zetaab . 

This PR should be merged so that there's no confusion about default config. Today the config is still deployed in the wrong spot making the setting False,[ yet the helm chart says it's true](https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/values.yaml#L1734) - WTF?

zetaab:
@chipzzz you are right.. I am currently using own fork of this helm chart, just fixing this one issue. It seems that others still uses broken config

chipzzz:
@zetaab , just wondering about your setup.. Are you running multiple/replicas - multiple shards? I overwrote this setting to true in actual CH originally then found problems with migrations( may have been due to compression set to true, need to test individually).

Currently I have both correctly set to false `compression` and `internal_replication` seems to work better but still trying to figure out correct fixes for add-node option https://github.com/getsentry/snuba/issues/6191

zetaab:
I have 3 shards with replicas. So in total 6 pods currently.

Mokto:
Still waiting for a proper migration path so that we don't break anything for existing users.

zetaab:
@Mokto could you explain why you need migration path? Currently the replication is not working at all because the configuration is placed to wrong place in configuration and it does not do anything. Which means that the behaviour is then default behaviour.

chipzzz:
To add there is 2 configs for this 1 in clickhouse and in in sentry

https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/values.yaml#L1734

https://github.com/sentry-kubernetes/charts/blob/develop/charts/clickhouse/values.yaml#L302

sentry/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 20.5.3
+version: 20.5.4
```
Mokto, line 5: Can you please remove that change ?

clickhouse/Chart.yaml
```diff
@@ -11,6 +11,6 @@ keywords:
 name: clickhouse
 sources:
 - https://github.com/sentry-kubernetes/charts
-version: 3.5.0
+version: 3.5.1
```
Mokto, line 14: same here

sentry/Chart.yaml
chipzzz: @zetaab

clickhouse/Chart.yaml
chipzzz: @zetaab

Mokto:

PR #920, opened 2023-06-28
Add existing secret configuration for Google SSO

lazy-slackoth:
Now there is no possibility to pass clientId and clientSecret for [Google SSO config](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L811-L814) from existing secret.
This PR aimed to fix this.


sentry/templates/configmap-sentry.yaml
```diff
@@ -55,10 +55,10 @@ data:
     ##########
     # Google #
     ##########
-    {{- if .Values.google.clientId }}
-    auth-google.client-id: {{ .Values.google.clientId | quote }}
-    auth-google.client-secret: {{ .Values.google.clientSecret | quote }}
-    {{ end }}
+    {{- if and (.Values.google.clientId) (.Values.google.clientSecret) (not .Values.google.existingSecret) }}
```
Mokto, line 58: I think you can make it `{{- if and (.Values.google.clientId) (.Values.google.clientSecret) }}` that's enough.

We should pass either of the possibilities

sentry/templates/_helper.tpl
```diff
@@ -533,6 +533,18 @@ Common Sentry environment variables
       name: {{ .Values.github.existingSecret }}
       key: {{ default "client-secret" .Values.github.existingSecretClientSecretKey }}
 {{- end }}
+{{- if and .Values.google.existingSecret }}
```
Mokto:
```
{{- if .Values.google.existingSecret }}
```

sentry/templates/configmap-sentry.yaml
lazy-slackoth: It's just additional step to protect against case where user define _.Values.google.clientId_ and _.Values.google.existingSecret_.

vjdhama:
We are also looking for this.

Is there something I can help with to get this merged?

lazy-slackoth:
Honestly saying I don't understand why MR still in open state, @Mokto could you prompt what else can be fixed?

sentry/templates/configmap-sentry.yaml
```diff
@@ -55,10 +55,10 @@ data:
...
+    auth-google.client-id: {{ .Values.google.clientId }}
+    auth-google.client-secret: {{ .Values.google.clientSecret }}
```
Mokto: Why did you remove `quote` here?

lazy-slackoth: Good that you noticed that, will return

Mokto (changes requested):
Please upgrade the minor version as you're adding a new feature.

Mokto (approved):

Mokto:
Thanks!

Issue #918, opened 2023-06-28
The Session Replay menu is not visible.


hyunkj42:
Hello.
I upgraded to sentry 23.3.1. (18.0.0)
The relay, snuba-subscription-consumer-events, and snuba-subscription-consumer-transactions pods have errors, but the web is normally started.
However, the expected **Session Replay** menu is not visible.
![image](https://github.com/sentry-kubernetes/charts/assets/22021000/4c5011c8-0162-405f-b5d5-db94e546b7cb)

expected Session Replay menu ㅜ
![image](https://github.com/sentry-kubernetes/charts/assets/22021000/243e8e0f-209e-4a76-b37c-8f67279183fb)

Can you tell me why?

hiloyt:
Because this is not a default option.

If you would like to see this menu, you need to add:

`"organizations:session-replay-ui"`

to the `templates/configmap-sentry.yaml` in the `SENTRY_FEATURES.update(` section.

hyunkj42:
@hiloyt 
oh thank you :)
After adding the settings you mentioned, the menu will appear.
thank you.

hyunkj42:
@hiloyt 
But why is that value missing from configmap-sentry.yaml in the helm chart provided here?

I made the session replay menu appear, but does it work normally???
In fact, I haven't tested it yet because I don't know how to test it. ;;

hiloyt:
As I mentioned before, this option is set to `false` by default in the server.py for Sentry self-hosted. So I think the authors of this chart decide to do this as well (disabled by default). You still can find Replays in the projects, you don't need this sidebar link.

In my company, we are using this feature and works great. Of course, you need also to edit your apps to support this feature, not only enable it for the Sentry instance.

https://docs.sentry.io/product/session-replay/getting-started/?original_referrer=https%3A%2F%2Fwww.google.co.uk%2F

bcarroll22:
@hiloyt I'm not sure this is true anymore.

https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L253

https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/configmap-sentry.yaml#L357

This repo doesn't seem to handle features the same way that self hosted does. This seems to overwrite the whole features array on line 357, whereas the self hosted only updates specific values. If you run a fresh install of the self-hosted version you'll have session replay and the menu using their features array, but not so with the helm version.

hiloyt:
[...]

PR #923, opened 2023-06-30
Update configmap-sentry.yaml

bcarroll22:
Per discussion with @hiloyt and @hyunkj42 in [this issue](https://github.com/sentry-kubernetes/charts/issues/918) I took a stab at updating the Sentry web config [to match what the Sentry self-hosted repo does currently](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py). I tried to preserve what looked to be intentional tweaks in this file, while also adding a few things that seemed to be missing in order to make it match current state of self-hosted.

If this isn't the right way to make the change, or doesn't match what the team would like to do, open to that feedback... just wanted to get it open for discussion in hopes it might make it easier to keep this repo in sync with what self-hosted does moving forward. The big driver behind this is that it seems self-hosted attempts to specifically override the value of _certain_ features in their config file, whereas this project seems to be trying to explicitly set the value of _all_ features. In the case of the discussion referenced above, this meant the `session-replay-ui` feature was missed in this repo's config, where it was enabled in the self-hosted repo. This was also causing the Profiling menu link being missing in the sidebar, so it's not a "one-off" problem (I don't know of any others off the top of my head, my team only encountered these two).

If it was desirable to have a smaller scope of change and not try to update to match current self-hosted, I believe removing [lines 357-359](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/templates/configmap-sentry.yaml#L357) would accomplish the same thing in a more focused way. Initializing features to an array with only `"auth:register"`, I believe, is the root cause of this particular issue reported above.

Let me know if there's any concerns or tweaks I should make, and thanks for maintaining this repo for the community!

### DIFF
--- a/sentry/templates/configmap-sentry.yaml
+++ b/sentry/templates/configmap-sentry.yaml
@@ -28,50 +28,18 @@ data:
       url: "http://{{ template "sentry.fullname" . }}-symbolicator:{{ template "symbolicator.port" }}"
     {{- end }}
 
-    ##########
-    # Github #
-    ##########
-    {{- with .Values.github.appId }}
-    github-app.id: {{ . }}
-    {{- end }}
-    {{- with .Values.github.appName }}
-    github-app.name: {{ . | quote }}
-    {{- end }}
-    {{- if not .Values.github.existingSecret }}
-      {{- with .Values.github.privateKey }}
-    github-app.private-key: {{- . | toYaml | indent 4 }}
-      {{- end }}
-      {{- with .Values.github.webhookSecret }}
-    github-app.webhook-secret: {{ . | quote }}
-      {{- end }}
-      {{- with .Values.github.clientId }}
-    github-app.client-id: {{ . | quote }}
-      {{- end }}
-      {{- with .Values.github.clientSecret }}
-    github-app.client-secret: {{ . | quote }}
-      {{- end }}
-    {{- end }}
-
-    ##########
-    # Google #
-    ##########
-    {{- if .Values.google.clientId }}
-    auth-google.client-id: {{ .Values.google.clientId | quote }}
-    auth-google.client-secret: {{ .Values.google.clientSecret | quote }}
-    {{ end }}
-
-    #########
-    # Slack #
-    #########
-    {{- if and (.Values.slack.clientId) (.Values.slack.clientSecret) (.Values.slack.signingSecret) (not .Values.slack.existingSecret) }}
-    slack.client-id: {{ .Values.slack.clientId | quote }}
-    slack.client-secret: {{ .Values.slack.clientSecret | quote }}
-    slack.signing-secret: {{ .Values.slack.signingSecret | quote }}
-    {{ end }}
-
-    #########
-    # Redis #
-    #########
+    ###################
+    # System Settings #
+    ###################
+    
+    # If this file ever becomes compromised, it's important to generate a new key.
+    # Changing this value will result in all current sessions being invalidated.
+    # A new key can be generated with `$ sentry config generate-secret-key`
+    system.secret-key: env('SENTRY_SECRET_KEY')
+    
+    # The ``redis.clusters`` setting is used, unsurprisingly, to configure Redis
+    # clusters. These clusters can be then referred to by name when configuring
+    # backends such as the cache, digests, or TSDB backend.
     redis.clusters:
       default:
         hosts:
@@ -85,6 +53,7 @@ data:
     ################
     # File storage #
     ################
+    
     # Uploaded media uses these `filestore` settings. The available
     # backends are either `filesystem` or `s3`.
     filestore.backend: {{ .Values.filestore.backend | quote }}
@@ -126,6 +95,50 @@ data:
       {{- if .Values.filestore.s3.location }}
       location: {{ .Values.filestore.s3.location | quote }}
       {{- end }}
+    {{ end }}
+
+    ######################
+    # GitHub Integration #
+    ######################
+    
+    {{- with .Values.github.appId }}
+    github-app.id: {{ . }}
+    {{- end }}
+    {{- with .Values.github.appName }}
+    github-app.name: {{ . | quote }}
+    {{- end }}
+    {{- if not .Values.github.existingSecret }}
+      {{- with .Values.github.privateKey }}
+    github-app.private-key: {{- . | toYaml | indent 4 }}
+      {{- end }}
+      {{- with .Values.github.webhookSecret }}
+    github-app.webhook-secret: {{ . | quote }}
+      {{- end }}
+      {{- with .Values.github.clientId }}
+    github-app.client-id: {{ . | quote }}
+      {{- end }}
+      {{- with .Values.github.clientSecret }}
+    github-app.client-secret: {{ . | quote }}
+      {{- end }}
+    {{- end }}
+
+    ######################
+    # Google Integration #
+    ######################
+    
+    {{- if .Values.google.clientId }}
+    auth-google.client-id: {{ .Values.google.clientId | quote }}
+    auth-google.client-secret: {{ .Values.google.clientSecret | quote }}
+    {{ end }}
+
+    #####################
+    # Slack Integration #
+    #####################
+    
+    {{- if and (.Values.slack.clientId) (.Values.slack.clientSecret) (.Values.slack.signingSecret) (not .Values.slack.existingSecret) }}
+    slack.client-id: {{ .Values.slack.clientId | quote }}
+    slack.client-secret: {{ .Values.slack.clientSecret | quote }}
+    slack.signing-secret: {{ .Values.slack.signingSecret | quote }}
     {{ end }}
 
     {{- if .Values.config.configYml }}
@@ -186,13 +199,6 @@ data:
     ###########
     # General #
     ###########
-
-
-    secret_key = env('SENTRY_SECRET_KEY')
-    if not secret_key:
-      raise Exception('Error: SENTRY_SECRET_KEY is undefined')
-
-    SENTRY_OPTIONS['system.secret-key'] = secret_key
 
     # Instruct Sentry that this install intends to be run by a single organization
     # and thus various UI optimizations should be enabled.
--- a/sentry/templates/configmap-sentry.yaml
+++ b/sentry/templates/configmap-sentry.yaml
@@ -19,15 +19,6 @@ data:
     system.url-prefix: {{ .Values.system.url | quote }}
     {{- end }}
 
-    # This URL will be used to tell Symbolicator where to obtain the Sentry source.
-    # See https://getsentry.github.io/symbolicator/api/
-    system.internal-url-prefix: 'http://{{ template "sentry.fullname" . }}-web:{{ .Values.service.externalPort }}'
-    symbolicator.enabled: {{ .Values.symbolicator.enabled }}
-    {{- if .Values.symbolicator.enabled }}
-    symbolicator.options:
-      url: "http://{{ template "sentry.fullname" . }}-symbolicator:{{ template "symbolicator.port" }}"
-    {{- end }}
-
     ###################
     # System Settings #
     ###################
@@ -96,6 +87,15 @@ data:
       location: {{ .Values.filestore.s3.location | quote }}
       {{- end }}
     {{ end }}
+
+    # This URL will be used to tell Symbolicator where to obtain the Sentry source.
+    # See https://getsentry.github.io/symbolicator/api/
+    system.internal-url-prefix: 'http://{{ template "sentry.fullname" . }}-web:{{ .Values.service.externalPort }}'
+    symbolicator.enabled: {{ .Values.symbolicator.enabled }}
+    {{- if .Values.symbolicator.enabled }}
+    symbolicator.options:
+      url: "http://{{ template "sentry.fullname" . }}-symbolicator:{{ template "symbolicator.port" }}"
+    {{- end }}
 
     ######################
     # GitHub Integration #
--- a/sentry/templates/configmap-sentry.yaml
+++ b/sentry/templates/configmap-sentry.yaml
@@ -360,9 +360,7 @@ data:
     ############
 
 
-    SENTRY_FEATURES = {
-      "auth:register": {{ .Values.auth.register | ternary "True" "False" }}
-    }
+    SENTRY_FEATURES["auth:register"] = {{ .Values.auth.register | ternary "True" "False" }}
     SENTRY_FEATURES["projects:sample-events"] = False
     SENTRY_FEATURES.update(
         {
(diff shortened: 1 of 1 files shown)


sentry/templates/configmap-sentry.yaml
```diff
@@ -304,35 +286,35 @@ data:
     ##############
 
     SENTRY_WEB_HOST = "0.0.0.0"
-    SENTRY_WEB_PORT = {{ template "sentry.port" }}
-    SENTRY_PUBLIC = {{ .Values.system.public | ternary "True" "False" }}
+    SENTRY_WEB_PORT = 9000
```
adonskoy, line 289: hardcoded value

sentry/templates/configmap-sentry.yaml
```diff
@@ -304,35 +286,35 @@ data:
...
     SENTRY_WEB_OPTIONS = {
         "http": "%s:%s" % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
         "protocol": "uwsgi",
-        # This is needed to prevent https://git.io/fj7Lw
+        # This is needed in order to prevent https://github.com/getsentry/sentry/blob/c6f9660e37fcd9c1bbda8ff4af1dcfd0442f5155/src/sentry/services/http.py#L70
         "uwsgi-socket": None,
+        "so-keepalive": True,
         # Keep this between 15s-75s as that's what Relay supports
-        "http-keepalive": {{ .Values.config.web.httpKeepalive }},
+        "http-keepalive": 15,
```
adonskoy, line 297: hardcoded value

adonskoy:
@bcarroll22, please double check your changes: you have hardcoded some parameters like web server port, http-keepalive and so on. Also, we can't just remove some configuration blocks like registration. the definition of this feature can be moved to SENTRY_FEATURES.update()

secret_key disappeared from the config and it's not clear why. If this was a copy-paste of sentry's default behavior, mention it in the PR description with a link to the same configuration in the sentry repository.

sentry/templates/configmap-sentry.yaml
```diff
@@ -28,50 +28,18 @@ data:
...
-    # Redis #
-    #########
+    ###################
+    # System Settings #
+    ###################
+    
+    # If this file ever becomes compromised, it's important to generate a new key.
+    # Changing this value will result in all current sessions being invalidated.
+    # A new key can be generated with `$ sentry config generate-secret-key`
+    system.secret-key: env('SENTRY_SECRET_KEY')
```
bcarroll22, line 29: moved from sentry.config.py

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/config.example.yml#L50)

sentry/templates/configmap-sentry.yaml
```diff
@@ -128,6 +88,59 @@ data:
       {{- end }}
     {{ end }}
 
+    # This URL will be used to tell Symbolicator where to obtain the Sentry source.
+    # See https://getsentry.github.io/symbolicator/api/
+    system.internal-url-prefix: 'http://{{ template "sentry.fullname" . }}-web:{{ .Values.service.externalPort }}'
```
bcarroll22, line 93: re-arranged

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/config.example.yml#L81)

sentry/templates/configmap-sentry.yaml
```diff
@@ -128,6 +88,59 @@ data:
...
+    symbolicator.enabled: {{ .Values.symbolicator.enabled }}
+    {{- if .Values.symbolicator.enabled }}
+    symbolicator.options:
+      url: "http://{{ template "sentry.fullname" . }}-symbolicator:{{ template "symbolicator.port" }}"
+    {{- end }}
+
+    ######################
```
bcarroll22, line 100: re-arranged

[reference](https://github.com/bcarroll22/charts/blob/develop/sentry/templates/configmap-sentry.yaml#L100)

sentry/templates/configmap-sentry.yaml
```diff
@@ -187,18 +179,26 @@ data:
...
+    )
+
+    #########
+    # Redis #
+    #########
+    
+    # Generic Redis configuration used as defaults for various things including:
+    # Buffers, Quotas, TSDB
+
+    SENTRY_OPTIONS["redis.clusters"] = {
```
bcarroll22, line 197: added redis config here to match

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L73)

sentry/templates/configmap-sentry.yaml
```diff
@@ -220,16 +220,20 @@ data:
...
-    #         "LOCATION": ["memcached:11211"],
-    #         "TIMEOUT": 3600,
-    #     }
-    # }
+    CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.memcached.MemcachedCache",
+            "LOCATION": ["{{ template "sentry.fullname" . }}-memcached:11211"],
+            "TIMEOUT": 3600,
+            "OPTIONS": {
```
bcarroll22, line 231: added additional param

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L117)

sentry/templates/configmap-sentry.yaml
```diff
@@ -304,35 +308,35 @@ data:
...
         "uwsgi-socket": None,
+        "so-keepalive": True,
         # Keep this between 15s-75s as that's what Relay supports
         "http-keepalive": {{ .Values.config.web.httpKeepalive }},
         "http-chunked-input": True,
         # the number of web workers
-        'workers': 3,
-        # Turn off memory reporting
+        "workers": 3,
+        "threads": 4,
```
bcarroll22, line 323: added additional config

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L208)

sentry/templates/configmap-sentry.yaml
```diff
@@ -304,35 +308,35 @@ data:
...
-    SENTRY_WEB_PORT = {{ template "sentry.port" }}
-    SENTRY_PUBLIC = {{ .Values.system.public | ternary "True" "False" }}
+    SENTRY_WEB_PORT = 9000
     SENTRY_WEB_OPTIONS = {
         "http": "%s:%s" % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
         "protocol": "uwsgi",
-        # This is needed to prevent https://git.io/fj7Lw
+        # This is needed in order to prevent https://github.com/getsentry/sentry/blob/c6f9660e37fcd9c1bbda8ff4af1dcfd0442f5155/src/sentry/services/http.py#L70
         "uwsgi-socket": None,
+        "so-keepalive": True,
```
bcarroll22, line 317: added additional config

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L202)

sentry/templates/configmap-sentry.yaml
```diff
@@ -352,113 +356,47 @@ data:
     ############
     # Features #
     ############
-
-
-    SENTRY_FEATURES = {
-      "auth:register": {{ .Values.auth.register | ternary "True" "False" }}
-    }
+
```
bcarroll22, line 359: primary change. updating certain values rather than overwriting the array

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L253)

sentry/templates/configmap-sentry.yaml
```diff
@@ -352,113 +356,47 @@ data:
...
-
-                "projects:plugins",
             )
         }
     )
 
     #######################
     # Email Configuration #
     #######################
+
```
bcarroll22, line 411: formatting for consistency

sentry/templates/configmap-sentry.yaml
```diff
@@ -468,51 +406,76 @@ data:
     SENTRY_OPTIONS['mail.host'] = os.getenv("SENTRY_EMAIL_HOST", {{ .Values.mail.host | quote }})
     SENTRY_OPTIONS['mail.from'] = os.getenv("SENTRY_EMAIL_FROM", {{ .Values.mail.from | quote }})
 
+    #######################
```
bcarroll22, line 421: added for geo integration

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L283)

sentry/templates/configmap-sentry.yaml
```diff
@@ -468,51 +406,76 @@ data:
...
+
+    {{- if .Values.geodata.path }}
+    GEOIP_PATH_MMDB = {{ .Values.geodata.path | quote }}
+    {{- end }}
+    
     #########################
     # Bitbucket Integration #
     #########################
-
+
```
bcarroll22, line 432: formatted for consistency

sentry/templates/configmap-sentry.yaml
```diff
@@ -468,51 +406,76 @@ data:
...
-    #########
-    # Relay #
-    #########
-    SENTRY_RELAY_WHITELIST_PK = []
-    SENTRY_RELAY_OPEN_REGISTRATION = True
-
-    #######################
-    # OpenAi Suggestions #
-    #######################
+    ##############################################
```
bcarroll22, line 436: updated to match self-hosted config for openai

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L296C1-L296C1)

sentry/templates/configmap-sentry.yaml
```diff
@@ -468,51 +406,76 @@ data:
...
+    # aligned with your company.
     
+    # Set the OPENAI_API_KEY on the .env or .env.custom file with a valid
+    # OpenAI API key to turn on the feature.
     OPENAI_API_KEY = os.getenv("OPENAI_API_KEY", "")
+    
     if OPENAI_API_KEY:
       SENTRY_FEATURES["organizations:open-ai-suggestion"] = True
+    
+    ##############################################
```
bcarroll22, line 451: new CSP settings from self-hosted

[reference](https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L311)

sentry/templates/configmap-sentry.yaml
```diff
@@ -468,51 +406,76 @@ data:
...
         'port': 9125,
     }
-{{- end }}
+    {{- end }}
 
-{{- if .Values.slack.existingSecret }}
+    {{- if .Values.slack.existingSecret }}
     #########
-    # SLACK #
+    # Slack #
```
bcarroll22, line 474: I left these in, but I think they may actually be duplicative because of the configuration in `config.yml` [(which is where it lives in self hosted)](https://github.com/getsentry/self-hosted/blob/master/sentry/config.example.yml#L109)

i can remove these if it's covered by having it in the other place

bcarroll22 (commented):
addressing comments

pavels:
Hi, just small note

These feature flags is going to be removed:
```
"organizations:metrics",
"organizations:metrics-extraction",
"organizations:transaction-metrics-extraction",
```

These are related to https://github.com/sentry-kubernetes/charts/pull/800

There was a bug in sentry, where metrics was not working properly without these feature flags (they have these on cloud version and forgot that they are not set on on-prem). The bug is fixed now but the helm chart already has all the infrastructure for `metrics-extraction` (topics are created, metrics services are created, tables in clickhouse exists). I don't know if sentry will eventually move forward with this approach or they will abandon it, but everything is in place for this now. Do you really want to drop this?

bcarroll22:
> Hi, just small note
> 
> These feature flags is going to be removed:
> 
> ```
> "organizations:metrics",
> "organizations:metrics-extraction",
> "organizations:transaction-metrics-extraction",
> ```
> 
> These are related to #800
> 
> There was a bug in sentry, where metrics was not working properly without these feature flags (they have these on cloud version and forgot that they are not set on on-prem). The bug is fixed now but the helm chart already has all the infrastructure for `metrics-extraction` (topics are created, metrics services are created, tables in clickhouse exists). I don't know if sentry will eventually move forward with this approach or they will abandon it, but everything is in place for this now. Do you really want to drop this?

I want to make sure I understand, you're saying the metrics don't work in self-hosted out of the box? If the feature is enabled on a fresh install of self-hosted, then this change shouldn't impact that. If it isn't enabled, I could add those lines back in with a comment that they're an intentional addition to the contents of the features array from self hosted.

pavels:
> > Hi, just small note
> > These feature flags is going to be removed:
> > ```
> > "organizations:metrics",
> > "organizations:metrics-extraction",
> > "organizations:transaction-metrics-extraction",
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > These are related to #800
> > There was a bug in sentry, where metrics was not working properly without these feature flags (they have these on cloud version and forgot that they are not set on on-prem). The bug is fixed now but the helm chart already has all the infrastructure for `metrics-extraction` (topics are created, metrics services are created, tables in clickhouse exists). I don't know if sentry will eventually move forward with this approach or they will abandon it, but everything is in place for this now. Do you really want to drop this?
> 
> I want to make sure I understand, you're saying the metrics don't work in self-hosted out of the box? If the feature is enabled on a fresh install of self-hosted, then this change shouldn't impact that. If it isn't enabled, I could add those lines back in with a comment that they're an intentional addition to the contents of the features array from self hosted.

Unfortunately this is more complicated. There is now 2 working paths for metrics collection - the "old path" and the "new path". Some time ago, when sentry team introduced the new path into their code, they removed the old path. By this, they broke the metrics on self-hosted and it was broken for i think 3 minor versions. I was hit by this bug (i use metrics to check if sentry processing pipeline works correctly as it really liked to brake on our installation) and decided to find a solution. I discovered the new path (which back then was completely missing from self-hosted and metrics was broken there) - it requires these 3 feature flags, 3 new services and small change in relay config. Metrics started to work. Some time later, sentry team decided that they want to fix the metrics on self-hosted. What they did is they put the old path back when these 3 feature flags are not set. So technically these are not needed now as well as the 3 additional services. The question is, is it good to revert back to the old path (to be inline with self-hosted) or keep the new one? This is for maintainers of this repo to decide - but if they decide that it should be reverted to the old path, than whole https://github.com/sentry-kubernetes/charts/pull/800 should be reverted, not just these 3 flags removed

bcarroll22:
ahh, thanks for the context, now i understand. 

in that case, i think for the purposes of this PR, rather than add scope i'd rather just add those lines back with a comment and it can be decided separately whether to revert #800 or continue down the current path.

sentry/templates/configmap-sentry.yaml
```diff
@@ -365,100 +366,49 @@ data:
...
-                "organizations:transaction-comparison",
-                "organizations:usage-stats-graph",
-                "organizations:inbox",
-                "organizations:unhandled-issue-flag",
-                "organizations:invite-members-rate-limits",
-                "organizations:dashboards-v2",
-                "organizations:reprocessing-v2",
+                "organizations:metric-alert-builder-aggregate",
+
+                # Leave these per discussion in https://github.com/sentry-kubernetes/charts/pull/800
```
bcarroll22, line 383: @pavels added this back to accommodate our conversation

Mokto:

PR #924, opened 2023-07-03
Splitting ingestConsumer into 3 services to be able to configure replicas and resources/limits granularly

pavelbrylov:
Splitting ingestConsumer into 3 services to be able to configure replicas and resources/limits granularly

Currently setttings are shared between 3 different services, this PR adds granularity to the configuration.


Mokto:
Thanks for your PR! 

I guess this is a breaking change and requires upgrading the major version.

pavelbrylov:
Should I bump it to 20, @Mokto ?
Also if we do it, maybe it's worth taking Kafka changes from #888 cause it works fine for us too (yes, one needs to recreate kafka volumes and there is a bit of data loss but it's worth it imho), as they would require major version change as well?

Mokto:
The automated deployment flow is not working well enough for us to be able to include both changes in 1 tag.

Let's do 20 with yours & then 21 with Kafka.

pavelbrylov:
and done

Mokto:
Could you also add a note in the README for the upgrade please ?

adonskoy:
How about changing the value structure a bit? Grouping settings of the same type of workloads looks more correct to me than a flatter one

Instead of:
``` yaml
sentry:
    ingestConsumerAttachments: {}
    ingestConsumerEvents: {}
    ingestConsumerTransactions: {}
```

It might be like this:
``` yaml
sentry:
    ingestConsumer:
        attachments: {}
        events: {}
        transactions: {}
```

PR #928, opened 2023-07-11
Add security context to clickhouse init container

pschwager:
in addition to [MR904](https://github.com/sentry-kubernetes/charts/pull/904) I also added an option to add the security context to the clickhouse init container


pschwager:
This PR is obsolete by #930

PR #929, opened 2023-07-11
add option to disable init container

pschwager:
I added an option to disable the init container as on my environment the container cannot be modified as the containers have to be running without root rights. Currently the used base image already has that directory configured, so we also don't need it. (Nevertheless leaving the default to enabled)

### DIFF
--- a/clickhouse/Chart.yaml
+++ b/clickhouse/Chart.yaml
@@ -11,6 +11,6 @@ keywords:
 name: clickhouse
 sources:
 - https://github.com/sentry-kubernetes/charts
-version: 3.3.0
+version: 3.5.0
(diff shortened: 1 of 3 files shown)


pschwager:
I just saw that someone else already had the same problem: [#609](https://github.com/sentry-kubernetes/charts/pull/609/files)

In this pull request @Mokto mentioned https://github.com/sentry-kubernetes/charts/pull/609#issuecomment-1152068773, that the init container could also be dropped completely. So if that's your preferred option, I can also adapt this PR and drop the init container

pschwager:
I added another PR to remove the init container completely. Feel free to decide, which one you like more... #930

PR #930, opened 2023-07-11
Drop useless clickhouse init container

pschwager:
So after a roundtrip I think we should drop the init container as the only thing it does is:
`mkdir -p /etc/clickhouse-server/metrica.d`

As the directory already exists in the base image, we don't need the mkdir.

So we also don't need to merge #928 #929 and can close them


Mokto:
Thanks! Could you please upgrade the minor version of clickhouse again?

pschwager:
@Mokto The branch was rebased!

Mokto:
Thanks!

Issue #902, opened 2023-06-05
not possible to run clickhouse without root


zetaab:
I am trying following

```
clickhouse:
  enabled: true
  clickhouse:
    podSecurityContext:
      fsGroup: 101
    securityContext:
      runAsUser: 101
      runAsNonRoot: true
      allowPrivilegeEscalation: false
```

This does not work with clickhouse.

Namespace is forced with baseline policy with label `pod-security.kubernetes.io/enforce: baseline`. (or enforce restricted)

The error message is

```
% kubectl logs sentry-clickhouse-1 -c init
mkdir: can't create directory '/etc/clickhouse-server/': Permission denied
```

Any tips how to get this running without root?

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #931, opened 2023-07-13
Upgrade clickhouse

pschwager:
I took a deeper look at the upgrade problem of clickhouse, as the old clickhouse version doesn't run as non-root in my environment. The trick was to configure the `sentry-clickhouse` cluster correct. After that the migration scripts will find the cluster and it's running

### DIFF
--- a/clickhouse/Chart.yaml
+++ b/clickhouse/Chart.yaml
@@ -11,6 +11,6 @@ keywords:
 name: clickhouse
 sources:
 - https://github.com/sentry-kubernetes/charts
-version: 3.3.0
+version: 3.4.0
 maintainers:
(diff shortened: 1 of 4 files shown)


zetaab:
@pschwager so does this solve https://github.com/sentry-kubernetes/charts/issues/902 ?

pschwager:
This one doesn't solve #902 directly, because currently the new clickhouse version doesn't work at all (even on root allowed envs), but I have another MR, which throws the init container away and that will solve your problem: #930

If I remember it correct, after you throw the init container away, there is another problem, which might hit you on a non-root env. This is then solved with this upgrade, too.

Mokto (approved):

PR #934, opened 2023-07-18
fix: Set Helm chart version of Clickhouse to 3.4.0 for Sentry

christiangonre:
After upgrading to version [19.5.0](https://github.com/sentry-kubernetes/charts/pull/931) the snuba migrations fails with the following error:

```
2023-07-18 13:04:25,668 Query: SELECT group, migration_id, status FROM migrations_dist FINAL WHERE group IN ('system', 'events', 'transactions', 'discover', 'outcomes', 'metrics', 'sessions', 'querylog', 'profiles', 'functions', 'replays', 'generic_metrics', 'test_migration', 'search_issues', 'spans')
2023-07-18 13:04:25,668 Block "" send time: 0.000036
Traceback (most recent call last):
  File "/usr/src/snuba/snuba/clickhouse/native.py", line 192, in execute
    result_data = query_execute()
  File "/usr/local/lib/python3.8/site-packages/clickhouse_driver/client.py", line 304, in execute
    rv = self.process_ordinary_query(
  File "/usr/local/lib/python3.8/site-packages/clickhouse_driver/client.py", line 491, in process_ordinary_query
    return self.receive_result(with_column_types=with_column_types,
  File "/usr/local/lib/python3.8/site-packages/clickhouse_driver/client.py", line 151, in receive_result
    return result.get_result()
  File "/usr/local/lib/python3.8/site-packages/clickhouse_driver/result.py", line 50, in get_result
    for packet in self.packet_generator:
  File "/usr/local/lib/python3.8/site-packages/clickhouse_driver/client.py", line 167, in packet_generator
    packet = self.receive_packet()
  File "/usr/local/lib/python3.8/site-packages/clickhouse_driver/client.py", line 184, in receive_packet
    raise packet.exception
clickhouse_driver.errors.ServerException: Code: 170.
DB::Exception: Requested cluster 'sentry-clickhouse' not found. Stack trace:
```

Looking at this PR, I can see that the Clickhouse version has been upgraded but Sentry haven´t been pointed to the new version, so the fix hasn´t been applied properly.

After pointing to the latest Clickhouse version, the deploy works without any problem.

PR #935, opened 2023-07-19
bump kafka and enable kraft mode by default

kimxogus:
This pr bump kafka chart to latest version and enable kraft mode by default.
It's a breaking change, so I bumped major version and added major version upgrade note

Also I added changes in #934 as latest sentry chart is broken for now. I'll merge latest develop again when #934 is merged.


Mokto:
Thanks! Although it's duplicate of https://github.com/sentry-kubernetes/charts/pull/888 that I just merged.

Kraft is not enabled per default, see the conversation in the PR.

PR #939, opened 2023-07-26
Sentry July release

adonskoy:


radubontila:
Check snuba-api probes, the health check url is not valid anymore in this version. You need to set **/health**.

Mokto (approved):

PR #940, opened 2023-07-26
Add ingest consumers HPA templates

dipalparmar:
Autoscaling for ingest consumers is mentioned in [values.yaml](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml#L174) but HPA's are not created.


patsevanton:
If scale ingest consumers automatically, you also need to change the number of partitions in a Kafka topic. This is not automatically possible.

Mokto:

Issue #932, opened 2023-07-13
Missing metrics on Kafka


psalkowski:
I have using latest chart with embeded Kafka and it looks like it misses two metrics:

- ingest-performance-metrics
- ingest-metrics

I have created them manually by logging into kafka pod and run:
```
kafka-topics.sh --create --topic ingest-metrics --bootstrap-server localhost:9092
kafka-topics.sh --create --topic ingest-performance-metrics --bootstrap-server localhost:9092
```

But that's not the best solution.

psalkowski:
It looks like Sentry have lots of pods with consume lot of resources during startup. In that case, helm hooks takes long time, but they are killed when exceed 100 seconds.

If anyone encounter this issue, the first thing that should be done is increasing hooks deadline by adding snippet into values.yaml:

```
          hooks:
            activeDeadlineSeconds: 300
```

Maybe it could be increased by default in this repo, as it doesn't harm anyone.

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #941, opened 2023-07-27
Longer hook timeout

artemisart:
From #932, to avoid killing hooks on slow startups.


patsevanton:
Fixed in v25.2.0. Please close the pull request

PR #954, opened 2023-08-11
Provision kafka topics in kafka-provisioning hook

kimxogus:
Disable kafka topic creation in `snuba-init` hook and move them to kafka-provisioning hook.
This will make kafka topic's partition, replication factor and other topic configs configurable as [here](https://github.com/bitnami/charts/blob/main/bitnami/kafka/values.yaml#L2014-L2026)

- ref: https://github.com/getsentry/snuba/issues/2188#issuecomment-975189971

### DIFF
--- a/sentry/values.yaml
+++ b/sentry/values.yaml
@@ -824,7 +824,7 @@ hooks:
   snubaInit:
     # As snubaInit doesn't support configuring partition and replication factor, you can disable snubaInit's kafka topic creation by setting `kafka.enabled` to `false`,
     # and create the topics using `kafka.provisioning.topics` with the desired partition and replication factor.
-    # Note that when you set `kafka.enabled` to `false`, snuba component might fail to start if brand new topics are not created by `kafka.provisioning`.
+    # Note that when you set `kafka.enabled` to `false`, snuba component might fail to start if newly added topics are not created by `kafka.provisioning`.
     kafka:
       enabled: true
     # podLabels: []
--- a/sentry/values.yaml
+++ b/sentry/values.yaml
@@ -822,6 +822,8 @@ hooks:
     # volumes: []
     # volumeMounts: []
   snubaInit:
+    kafka:
+      enabled: true
     # podLabels: []
     podAnnotations: {}
     resources:
--- a/sentry/values.yaml
+++ b/sentry/values.yaml
@@ -1164,6 +1164,51 @@ kafka:
   provisioning:
     enabled: true
     topics:
+      - name: events
+      - name: event-replacements
+      - name: snuba-commit-log
+      - name: cdc
+      - name: transactions
+      - name: snuba-transactions-commit-log
+      - name: snuba-metrics
+      - name: outcomes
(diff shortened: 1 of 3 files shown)


z0rc:
This will break use case when in-chart kafka is disabled and external kafka is used. This change requires at least documentation, at best this whole approach should be refactored to have helm hook or init container to provision topics independently from used kafka solution.

z0rc:
Plus this raises a question how this list of topics should be maintained. Where is the source of truth? How verify topic changes on next sentry release?

kimxogus:
@z0rc I added `hooks.snubaInit.kafka.enabled` and set `true` by default.
This won't break current behavior.

Mokto (approved):

Mokto:
@z0rc are you now fine with the change?

z0rc:
@Mokto not really. Current approach maintains backward compatibility, but doesn't add any documentation or single comment line.

> at least documentation, at best this whole approach should be refactored

Wasn't addressed at all.

> Plus this raises a question how this list of topics should be maintained. Where is the source of truth? How verify topic changes on next sentry release?

Not addressed at all.

adonskoy:
I see no reason to make this option enabled by default since we already have bootstrap enabled in kafka chart. It's possible to add this option by default for snuba migrations if external kafka is used.

adonskoy:
also in [getsentry/self-hosted](https://github.com/getsentry/self-hosted/blob/23.7.2/install/create-kafka-topics.sh#L6) there is a script that creates the "necessary" topics. If these topics are created by a snuba with this parameter, then our "source of truth" will be a snuba and we will not need to have our own list of topics

kimxogus:
@z0rc  Sorry I missed comments, I initially made default topic provisioning list from snuba-db-init job's log, and I found [this topic list](https://github.com/getsentry/snuba/blob/9d546c467792df43b2fad25b6d313698ff7ba5b9/snuba/utils/streams/topics.py#L6-L63), so I noted that in comment.
We can verify topic changes on next sentry release based on that topic enum file.
Is that ok?

sentry/values.yaml
```diff
@@ -1165,6 +1168,8 @@ kafka:
   socketRequestMaxBytes: "50000000"
   provisioning:
     enabled: true
+    # Topic list is based on https://github.com/getsentry/snuba/blob/master/snuba/utils/streams/topics.py
```
z0rc: There are additional topics at https://github.com/getsentry/self-hosted/blob/master/install/create-kafka-topics.sh#L6

z0rc (changes requested):

sentry/values.yaml
kimxogus: Added that in https://github.com/sentry-kubernetes/charts/pull/954/commits/4ce08e49e3688d519964c39e511989eb8c024bd2

z0rc (approved):

Mokto (approved):
Thanks guys

PR #958, opened 2023-08-14
fix: Create missing service accounts

christiangonre:
When you try to enable the profiling option with the `serviceAccount.enabled as true`, it fails because the vroom service account is not available.
It throw the following error:
` Warning  FailedCreate  29s (x17 over 5m58s)  replicaset-controller  Error creating: pods "sentry-vroom-fd8b6496c-" is forbidden: error looking up service account sentry/sentry-vroom: serviceaccount "sentry-vroom" not found`

Something similar happens with the cleanup cronjob service account, when you enable the `serviceAccount.enabled as true` it should create a service account for the cleanup cronjob, but this doesn´t happen.
I modified the code of the cleanup cronjob to match the same format of the rest of deployments, this is needed to have a better manage of the service account privileges

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 20.3.1
+version: 20.3.2
 appVersion: 23.7.2
 dependencies:
   - name: memcached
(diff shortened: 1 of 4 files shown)


Mokto (approved):

Mokto:
Thanks!

PR #966, opened 2023-08-16
fix: Fix wrong service account for ingest-profiles deployment

christiangonre:
The service account of the ingest-profiles deployment is wrong, this PR fixes that problem and points the service account to the right one.


Mokto:
Thanks! Could you please rebase your code?

christiangonre:
Done!

Mokto (approved):

PR #984, opened 2023-08-28
Renovate: update only charts version

adonskoy:
I've added an extra rule for Helm to limit the datasource that renovate will check.


Mokto:
Thanks.

Issue #1016, opened 2023-09-15
All events tab in issue breaks with Nginx error due to buffer size


gersmann:
There is a problem with Sentry, which requires larger buffers on both, the included Nginx service and the Nginx ingress. 

i managed to fix this on the ingress, by adding these annotations:
```
      nginx.ingress.kubernetes.io/proxy-buffer-size: 256k
      nginx.ingress.kubernetes.io/proxy-buffering: 'off'
      nginx.ingress.kubernetes.io/proxy-buffers-number: '4'
```

But this needs to be fixed in the included Nginx config as well, I don't know how. 

The issue: https://github.com/getsentry/self-hosted/issues/1927
And the PR which fixes it: https://github.com/getsentry/self-hosted/pull/1984

Is there a way to implement this with this chart?

Glaaj:
Isn't this related to the values provided to the nginx sub-chart?

I can see that bitnami nginx [chart](https://github.com/bitnami/charts/tree/main/bitnami/nginx) provides the following option: 

> existingServerBlockConfigmap: 
> "ConfigMap with custom server block to be added to NGINX configuration"

In there a reference to a configmap is set. I've edited the template to include the fix you mentioned. Pr here: [PR](https://github.com/sentry-kubernetes/charts/pull/1022)

Glaaj:
I've gone ahead and created a PR with the fix you linked from [getsentry/self-hosted](https://github.com/getsentry/self-hosted/pull/1984)

gersmann:
Wow, nice, thank you.

Mokto:
This issue is stale because it has been open for 30 days with no activity.

PR #1022, opened 2023-09-21
Add proxy buffer size config to fix Bad Gateway

Glaaj:
See: https://github.com/sentry-kubernetes/charts/issues/1016

Based on: https://github.com/getsentry/self-hosted/pull/1984


fidr (approved):

Mokto (approved):
Thanks!

PR #1035, opened 2023-10-12
Allow to use specified key for slack existing secrets

NikonNLG:
Allow to use predefined keys for slack client id, client secret and signing secret


sentry/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 20.7.0
+version: 20.7.1
```
ushuz: Need to bump the version again.

ushuz (approved):

sentry/Chart.yaml
NikonNLG: @ushuz Done

Mokto (approved):

PR #1054, opened 2023-10-30
fix: replay session

TartanLeGrand:
Add all configuration for replay session and let enable by default

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 20.8.3
+version: 20.9.1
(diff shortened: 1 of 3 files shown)


acjohnson:
+1 but does this actually make Session Replays "just work"? We get a blank "video" when we try to watch the session replay back... Maybe we are doing something wrong still...

TartanLeGrand:
> +1 but does this actually make Session Replays "just work"? We get a blank "video" when we try to watch the session replay back... Maybe we are doing something wrong still...

We get blank too, I start to add all features and will see next :smile:

Mokto:
Thanks. Could you please update to 20.8.3?

TartanLeGrand:
@Mokto good 😄

PR #1060, opened 2023-11-06
fix: nginx static path not pass proxy

gunslinger23:
`_static` path will not pass proxy which is not expected

### DIFF
--- a/sentry/templates/configmap-nginx.yaml
+++ b/sentry/templates/configmap-nginx.yaml
@@ -42,6 +42,7 @@ data:
       }
 
       location /_static/ {
+        proxy_pass http://sentry;
         proxy_hide_header Content-Disposition;
       }
     }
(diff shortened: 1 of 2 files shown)


gunslinger23:
ci lint seem not right.

Mokto:
Because you have to bump the version of the chart

gunslinger23:
fixed

sentry/templates/configmap-nginx.yaml
```diff
@@ -42,6 +42,7 @@ data:
       }
 
       location /_static/ {
+        proxy_pass http://sentry;
```
Mokto, line 45: I think that would only work if the chart name (and therefore the service name) is "sentry"

adonskoy: This will work. This is the name of the [upstream](https://github.com/sentry-kubernetes/charts/blob/0b9d7e89768745d1b21091cc0be1b778a157d6a9/sentry/templates/configmap-nginx.yaml#L12), not the service

Mokto:
Thanks.

PR #1065, opened 2023-11-14
Support HPA container resource

kimxogus:
Autoscaling v2 has been GA since k8s 1.23.
Container resource is enabled if k8s version is >= 1.27


adonskoy:
It seems to me that this change needs to be released as a new major version, and also in the README, indicate the minimum version of kubernetes with which this chart can be used.

You can also add kubeVersion to Chart.yaml. You can specify `kubeVersion: >=1.23`

kimxogus:
@adonskoy I added backward compatibility for k8s < 1.23 because I think this change is not big enough to bump a major version.

kimxogus:
@adonskoy Could you review this?

ushuz (approved):

kimxogus:
@Mokto Could you merge this?

Mokto (approved):

PR #1080, opened 2023-11-24
feat: Allow to set max_suspicious_broken_parts in merge_tree settings…

mrouhi13:
A few days ago I faced a problem in one of our clickhouse replicas. When clickhouse tries to init db it needs to merge broken parts of data to a single partition and it needs to delete those files afterward. If the number of those files exceeded the number set in the configuration it denied deletion ([here](https://clickhouse.com/docs/en/operations/settings/merge-tree-settings)) and db doesn't init, so what we need to do is increase the max_suspicious_broken_parts value to cover those broken file numbers.

Unfortunately, it is not possible to set the above settings via chart values so I decided to add handling these specific params.

### DIFF
--- a/clickhouse/templates/configmap-config.yaml
+++ b/clickhouse/templates/configmap-config.yaml
@@ -165,6 +165,7 @@ data:
             <parts_to_delay_insert>{{ .Values.clickhouse.configmap.merge_tree.parts_to_delay_insert }}</parts_to_delay_insert> 
             <parts_to_throw_insert>{{ .Values.clickhouse.configmap.merge_tree.parts_to_throw_insert }}</parts_to_throw_insert> 
             <max_part_loading_threads>{{ .Values.clickhouse.configmap.merge_tree.max_part_loading_threads }}</max_part_loading_threads> 
+            <max_suspicious_broken_parts>100</max_suspicious_broken_parts>
         </merge_tree>
         {{- end }}
     </yandex>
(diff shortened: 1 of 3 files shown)


Mokto (approved):
Thanks. Could you please upgrade the minor version of the chart instead?

clickhouse/templates/configmap-config.yaml
```diff
@@ -165,6 +165,7 @@ data:
             <parts_to_delay_insert>{{ .Values.clickhouse.configmap.merge_tree.parts_to_delay_insert }}</parts_to_delay_insert> 
             <parts_to_throw_insert>{{ .Values.clickhouse.configmap.merge_tree.parts_to_throw_insert }}</parts_to_throw_insert> 
             <max_part_loading_threads>{{ .Values.clickhouse.configmap.merge_tree.max_part_loading_threads }}</max_part_loading_threads> 
+            <max_suspicious_broken_parts>100</max_suspicious_broken_parts>
         </merge_tree>
```
mzglinski, line 169: This shouldn't be hardcoded.

mrouhi13: You're right, I will fix it.

DanielMcAssey: Possible to push a change for this?

mrouhi13: PR created.

PR #1106, opened 2024-01-06
Update sentry to 23.11.2

xvilo:
Upgrading to last 23.11 version, currently not updating to 23.12 on purpose

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 20.12.1
+version: 20.12.2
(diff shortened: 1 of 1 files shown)


Mokto (approved):

xvilo:
Deploy this on our cluster and seems to run fine so far, thanks @Mokto 👍

kfkawalec:
What is the reason for not upgrading to 23.12?

xvilo:
Because I'm unsure about the procedures and possibly new stuffs needed as 23.11.x seems to be a minor/patch release in Sentrys versioning scheme. I do not have an easy test set-up so that makes a bit harder for me. I'd rather have people do it that have experience with bigger/major version upgrades

PR #1131, opened 2024-02-05
fix(memcache): do not set SERVER_MAX_VALUE_LENGTH with latest django versions

adonskoy:
Memcache module is no longer present in the Django package. Leaving this in the configuration will cause startup failures


adonskoy:
@Mokto, FYI

Mokto (approved):
Thanks!

Issue #1132, opened 2024-02-06
How to enable User Feedback feature 


ajith-nataraj:
How can we enable user-feedback for sentry self-hosted via helm charts

 SENTRY_FEATURES['organizations:issue-feedback'] = True
 tried this in configmap-sentry.yaml and it did not work
 
 tried the below also did not work
   features:
    enableProfiling: true   #Enable Profiling
    enableSessionReplay: true #Enable Session Replay
    enableUserFeedback: true  #Enable User Feedback
    orgSubdomains: true
    vstsLimitedScopes: true
 
 
can you please help on how we can enable User Feedback feature  as its done on self-hosted sentry 

Any Help would be very much appreciated 
 https://github.com/getsentry/self-hosted/issues/2646

mihallis:
#1142

ajith-nataraj:
<img width="1289" alt="Screenshot 2024-02-14 at 12 44 11" src="https://github.com/sentry-kubernetes/charts/assets/107095805/19323fab-52e6-4f41-ab5e-283ea4ed3a1e">
Getting this error after enabling the feature, anything we are missing here 
please help

ajith-nataraj:
@CMDMichalKoval  did you face the above issue after enabling User-feedback ?

mihallis:
i'll check it

@ajith-nataraj can you send list of features in sentry-sentry configmap?

ajith-nataraj:
SENTRY_FEATURES["projects:sample-events"] = False
    SENTRY_FEATURES.update(
        {
            feature: True
            for feature in ("organizations:advanced-search",
                organizations:android-mappings",
                organizations:api-keys",
                organizations:boolean-search",
                organizations:related-events",
                organizations:alert-filters",
                organizations:custom-symbol-sources",
                organizations:dashboards-basic",
                organizations:dashboards-edit",
                organizations:data-forwarding",
                organizations:discover",
                organizations:discover-basic",
                organizations:discover-query",
                organizations:discover-frontend-use-events-endpoint",
                organizations:enterprise-perf",
                organizations:event-attachments",
                organizations:events",
                organizations:global-views",
                organizations:incidents",
                organizations:metric-alert-builder-aggregate",
                organizations:metric-alert-gui-filters",
                organizations:integrations-event-hooks",
                organizations:integrations-issue-basic",
                organizations:integrations-issue-sync",
                organizations:integrations-alert-rule",
                organizations:integrations-chat-unfurl",
[...]

PR #1142, opened 2024-02-10
add option to enable feedback ui

mihallis:
Add option to enable feedback ui

```yaml
  features:
    orgSubdomains: false
    vstsLimitedScopes: true
    enableProfiling: false
    enableSessionReplay: true
    enableFeedback: false # <---
```

Inspired by: #1132

### DIFF
--- a/sentry/Chart.yaml
+++ b/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 21.3.3
+version: 21.3.4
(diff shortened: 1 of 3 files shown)


sentry/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 21.3.3
+version: 21.3.4
```
Mokto: Can you please upgrade the minor version?

mihallis: Updated to 21.4.0

PR #1154, opened 2024-02-20
upgrade sentry to 24.2.0

mihallis:


Mokto:
Hi, thanks for your PR.

1. Are you sure that's enough?
2. Have you tried it specifically?

mihallis:
Yes, i tried on few deployments. All dependencies looks compatible.

maths22:
Based on this diff, https://github.com/getsentry/self-hosted/compare/24.1.2...24.2.0 it wouldn't appear like there should be any other changes needed

PR #1174, opened 2024-03-13
allow to use liveness and healthcheck for ingest-consumer-transactions and snuba-transactions-consumer

pawelmrowka:
To prevent consumer hang-ups, for example, due to temporary unavailability of Kafka.

example usage:
```
sentry:
        ingestConsumer:
        healthCheckFilePath: /tmp/healthcheck
        livenessProbe:
            exec:
                command:
                    - rm
                    - /tmp/healthcheck
            initialDelaySeconds: 5
            periodSeconds: 320
 snuba:
     transactionsConsumer:
        healthCheckFile: /tmp/healthcheck
        livenessProbe:
            exec:
                command:
                    - rm
                    - /tmp/healthcheck
            initialDelaySeconds: 5
            periodSeconds: 320
```


pawelmrowka:

Mokto (commented):
Why wouldn't we apply that per default for everyone?

Mokto:
In that case I don't even think it's necessary that as a parameter in values.yaml

PR #1192, opened 2024-03-28
(clickhouse) Fix merge_tree section format in configmap-config.yaml

NikolaiIuzhaninYS:
Otherwise if merge_tree is enabled, the pods of StatefulSet will be in a crash-loop with Clickhouse unable to merge configs


NikolaiIuzhaninYS:
@Mokto Could you please review this so I could update the reference in the sentry main chart?

Mokto (approved):

Issue #1212, opened 2024-04-11
Help needed on migrating from kafka with zookeeper to Kraft


fayak:
Hello !

I'm looking to upgrade my deployment from 21.x to 22.x. I'm trying to follow the instructions from the README, but I have a hard time figuring a few things out.
On the step 3 :
```
Deploy at least one Kraft controller-only in your deployment with zookeeperMigrationMode=true. The Kraft controllers will migrate the data from your Kafka ZkBroker to Kraft mode.

To do this, add the following values to your Zookeeper deployment when upgrading:

controller:
    replicaCount: 1
    controllerOnly: true
    zookeeperMigrationMode: true
broker:
    zookeeperMigrationMode: true
kraft:
    enabled: true
    clusterId: "<your_cluster_id>"

```
What should be the value of `kafka.zookeeper.enabled` there ? If i understand correctly, we want to keep zookeeper at first to migrate things from it to kafka conroller, and once it's done, only then remove zookeeper.
If I set `kafka.zookeeper.enabled: true`, I have an error thrown
```
coalesce.go:237: warning: skipped value for kafka.config: Not a table.
Error: execution error at (sentry/charts/kafka/templates/NOTES.txt:333:4): 
VALUES VALIDATION:

kafka: Zookeeper mode - No Kafka brokers configured
    Zookeper mode has been enabled, but no Kafka brokers nodes have been configured
```

Setting it to false (or not setting it) looks like it's working, but I don't see a sts for zookeeper anymore. Is it expected ? I won't loose any data proceeding ?

@TartanLeGrand I think you wrote the migration guide, could you advise please ? :)

TartanLeGrand:
Hello @fayak 👋,

> What should be the value of kafka.zookeeper.enabled there ? If i understand correctly, we want to keep zookeeper at first to migrate things from it to kafka conroller, and once it's done, only then remove zookeeper.

Instead you dont have reach the step 6 according to the migration doc you have to keep `kafka.zookeeper.enabled: true`.

Why you set `kafka.zookeeper.enabled: true` you are not under zookeeper before it's an fresh install ?

fayak:
Hi !

> Instead you dont have reach the step 6 according to the migration doc you have to keep kafka.zookeeper.enabled: true.

Ok, I think this used to be the default value before, so I add it explicitly now, but therefore get the templating error I've mentioned.

Apparently I need to also add `broker.replicacount` :
```
kafka:
  broker:
[...]

PR #1218, opened 2024-04-13
When set to "controller only", port 9092 is not listened on.

pyama86:
This issue probably relates to [this problem](https://github.com/sentry-kubernetes/charts/issues/1212). When set to [controllerOnly](https://github.com/bitnami/charts/blob/94ce52ab08a89a7881e8d443f153e284a3f65229/bitnami/kafka/templates/controller-eligible/statefulset.yaml#L253), port 9092 is not listened on, so this change is necessary.

### DIFF
--- a/sentry/templates/_helper.tpl
+++ b/sentry/templates/_helper.tpl
@@ -411,6 +411,17 @@ Set Kafka Confluent port
 {{- end -}}
 
 {{/*
+Set Kafka Confluent Controller port
+*/}}
+{{- define "sentry.kafka.controller_port" -}}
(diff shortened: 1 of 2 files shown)


TartanLeGrand (approved):

TartanLeGrand:
@pyama86 Change your commit to an conventional commits => https://www.conventionalcommits.org/en/v1.0.0/ 😄

pyama86:
@TartanLeGrand Thanks!I fixed it.

Mokto:
Thanks ! Let's see how to the new release system works :)

PR #1222, opened 2024-04-16
fix kafka.listeners.interBroker typo

kimxogus:
https://github.com/bitnami/charts/blob/main/bitnami/kafka/values.yaml#L173


kimxogus:
@Mokto Can you review this too?

Mokto (approved):

PR #1228, opened 2024-04-19
feat: checksum only for configmap contents

kimxogus:
Checksum for `configmap-{COMPONENT}.yaml` file contains `chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"` label which makes checksum to be changed in every helm chart version bump.
This makes all the deployments rollout in every helm chart bump even there's no actual change in any manifests.

This change will make checksum changes only when actual config content is updated.


kimxogus:
@Mokto Could you review this PR?

Mokto (approved):

Mokto:
Thanks.

PR #1232, opened 2024-04-22
fix: nginx configuration for metrics

bkk-bcd:
Fixes to `configmap-nginx.yaml`:

* It's valid/applicable to enable the status endpoint when either `metrics.enabled` or `metrices.serviceMonitor.enabled`. We don't need the `serviceMonitor` to scrape metrics, that's an optional Prometheus component.
* We shouldn't include the status endpoint in access logs
* Be explicit about `stub_status` enablement


Mokto (approved):

bkk-bcd:
🍻

PR #1239, opened 2024-04-24
fix: sentry metrics deployment annotations

bkk-bcd:
Annotations for the web deployment pod should be separated from the annotations for the metrics deployment.


Mokto (approved):

bkk-bcd:
👍

PR #1240, opened 2024-04-27
Add liveness Check to consumers

pyhp2017:
Consumers needed a liveness check. In multiple scenarios, we noticed that consumers weren't able to continue working due to the lags in topics, but by restarting the pod it will be fixed without any manual intervention.

Sentry & Snuba already provide this functionality for consumers, but it wasn't implemented in Kubernetes manifests.


Mokto:
Thanks for your PR.

I don't think we need to allow people to disable the healthcheck probes.
Do you see any reason otherwise?

pyhp2017:
As you are aware, some of the consumers have the "auto-offset-reset" setting enabled. This means that at the start, they will reset to either the latest or earliest offset. However, this feature may not be necessary for everyone.

Mokto (approved):

PR #1245, opened 2024-05-08
feat: ingest consumers and workers separation

NikolaiIuzhaninYS:
These changes give an ability to define resource quotas and autoscaling separately for every ingest-consumer according to usage pattern. It allows to choose proper granularity for each consumer and to optimise resource consumption.

`helm upgrade --force` can be used to recreate Deployments (labels are immutable), as well as manual deletion of these Deployments before upgrade


Mokto:
So it's a breaking change in a way ?

NikolaiIuzhaninYS:
> So it's a breaking change in a way ?

Yes, it is. Does your question mean that it must be a major release and I should add a migration guide?

I also intend to make similar changes for workers: separate events and transactions processing deployments to improve flexibility there. Too big to combine with these changes but would be reasonable to put into a single chart release.

pavelbrylov:
https://github.com/sentry-kubernetes/charts/pull/924 Sounds familiar :) Feel free to close the old one once this is merged. I unfortunately didn't have time to apply the requested changes

NikolaiIuzhaninYS:
https://github.com/sentry-kubernetes/charts/pull/1251 contains the changes I mentioned above @Mokto

Mokto:
Thanks ! Can you please remove the formatting + add breaking changes notes to the readme ? And same in the other PR.
I'd also like to merge both your commits into a temporary branch I just created called `split-scaling`.

So that we send everything at once.

NikolaiIuzhaninYS:
@Mokto Done, also decided to combine changes in this single PR

Mokto:
Thanks!

PR #1254, opened 2024-05-14
fix: fix snuba-outcomes-billing-consumer args

kimxogus:
`--raw-events-topic` arg was missing.

https://github.com/getsentry/self-hosted/pull/2909/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R270

### DIFF
--- a/sentry/templates/deployment-snuba-outcomes-billing-consumer.yaml
+++ b/sentry/templates/deployment-snuba-outcomes-billing-consumer.yaml
@@ -80,6 +80,8 @@ spec:
           - "--consumer-group"
           - "snuba-consumers"
           - "--auto-offset-reset"
+          - "--raw-events-topic"
+          - "outcomes-billing"
           - "{{ .Values.snuba.outcomesBillingConsumer.autoOffsetReset }}"
           - "--max-batch-time-ms"
           - "750"


kimxogus:
@Mokto Can you review this?

sentry/templates/deployment-snuba-outcomes-billing-consumer.yaml
```diff
@@ -80,6 +80,8 @@ spec:
           - "--consumer-group"
           - "snuba-consumers"
           - "--auto-offset-reset"
+          - "--raw-events-topic"
+          - "outcomes-billing"
           - "{{ .Values.snuba.outcomesBillingConsumer.autoOffsetReset }}"
```
farodin91, line 85: This should be behind auto offset reset

rdziwinski:
Now there is another problem:
`Error: Invalid value for '--auto-offset-reset': '--raw-events-topic' is not one of 'error', 'earliest', 'latest'.`
Is it should be like below?

`--auto-offset-reset=earliest --raw-events-topic outcomes-billing`

farodin91:
https://github.com/sentry-kubernetes/charts/pull/1257

PR #1270, opened 2024-05-20
feat: bump sentry to 24.5.0

kimxogus:


Mokto:
Thanks! Have you tested the behavior on your cluster? Have you made sure the changes from https://github.com/getsentry/self-hosted are reflected here?

kimxogus:
@Mokto I've tested in my test environment with bumping sentry app version only.
[getsentry/self-hosted](https://github.com/getsentry/self-hosted/releases/tag/24.5.0) has clickhouse version upgrade, but it doesn't look necessary because it worked with our helm chart's default clickhouse version too.

PR #1275, opened 2024-05-21
fix(worker): workerTransactions should be disabled by default

NikolaiIuzhaninYS:


Mokto:
Why?

NikolaiIuzhaninYS:
> Why?

It was an initial idea to have "just worker" consuming all queues enabled by default, and additional "special workers" consuming particular queues to be enabled on demand (if you have performance transactions prevailing in ingestion, you could shift balance enabling additional workers for transactions-related queues, same  thing for errors)

PR #1278, opened 2024-05-22
feat: allow to configure extra manifests through values.yaml

karol-szymanowski:
This change allows users to deploy additional kubernetes manifests within this chart.

## Why?

By adding additional objects (like [external-secret](https://external-secrets.io/latest/)) I can mount secrets on which this app depends on, very useful when chart is deployed through tools like ArgoCD. It's generic enough that it will find many more use cases.

I was inspired by this thread: https://github.com/argoproj/argo-helm/pull/1094

### DIFF
--- a/sentry/README.md
+++ b/sentry/README.md
@@ -109,6 +109,7 @@ Note: this table is incomplete, so have a look at the values.yaml in case you mi
 | `snuba.transactionsConsumer.enabled`       | Enable Snuba transactions consumer                                                                                                                                | `true`                          |
 | `symbolicator.api.enabled`                    | Enable Symbolicator                                                                                                                                                 | `false`                        |
 | `symbolicator.api.config`                     | Config file for Symbolicator, see [its docs](https://getsentry.github.io/symbolicator/#configuration)                                                               | see values.yaml                |
+| `extraManifests`                              | Extra K8s manifests to deploy                                                                                                                                       | []                             |
 
 ## NGINX and/or Ingress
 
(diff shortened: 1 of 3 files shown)


Mokto:
I'm not sure I like the added complexity.

In my opinion it's better to create a new chart that has the sentry dependency where you can add as many manifests as you want.

karol-szymanowski:
I think the cost of adding this change is very low and it makes the chart much more flexible for different setups. In some cases creating separate helm chart means that you need to configure the whole CI/CD just to add one resource.

It's especially helpful when you use gitops and tools like ArgoCD, if I had to configure my own chart I would lose the ability to autoupdate the chart in argo. Instead I'd need some additional tools that would update helm dependencies or rely on some hacky solution.

From my experience pretty much every major helm chart let users to expand charts like that, examples:
- [oauth2proxy](https://github.com/oauth2-proxy/manifests/pull/76/files)
- [most if not all of bitnami helm charts](https://github.com/bitnami/charts/pull/6536/files)
- argocd
- [fluxcd](https://github.com/fluxcd-community/helm-charts/blob/main/charts/flux2/templates/extra-manifests.yaml)
- [grafana](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/extra-manifests.yaml)
- [atlantis](https://github.com/helm/charts/blob/master/stable/atlantis/templates/extra-manifests.yaml)

I don't say that this is the best practise but I think it became sort of a standard and not having it might put some people in uncomfortable spot. As I said, from the maintainer POV the cost of adding it should be minimal and it can reduce a lot of complexity on a user side.

Mokto:
I'm still not fund of the idea but you're right, it doesn't bring that much complexity!

PR #1279, opened 2024-05-22
feat: Allow to configure external postgres with values from secret

karol-szymanowski:
This change allows to configure external postgres with existing secret. The way it works is that all the variables for postgres connection config are loaded from the environment variables (now only password is). This makes a postgres configuration more flexible and opens up a possibility to load those variables from different sources, eg. secret, config map or directly in the deployment.

> ⚠️ This is a non breaking change, I made sure that `.Values.externalPostgresql.existingSecretKey` will still work but it should be noted that this property is depricated and `.Values.externalPostgresql.existingSecretKeys.password` should be used instead.

## Why?

In our case database provisioning is fully automated along with the secret that contains all the connection configuration. This change allows us to use existing secrets to configure external postgres.

### DIFF
--- a/sentry/README.md
+++ b/sentry/README.md
@@ -230,7 +230,46 @@ externalKafka:
     port: 9092
 ```
 
+## External Postgres configuration
 
+You can either pass postgres connection credentials directly in `values.yaml`:
+
+```yaml
+externalPostgresql:
+  host: postgres
+  port: 5432
+  username: postgres
+  password: postgres
+  database: sentry
+```
+
+or use existing `secret` like in the example below:
+
+```yaml
+externalPostgresql:
+  existingSecret: secret-name
+  existingSecretKeys:
+    password: password
+    username: username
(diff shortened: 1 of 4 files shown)


karol-szymanowski:
@Mokto could you review it?

Mokto (approved):

Mokto:
Thanks! Sorry about the delay. Could you please rebase your branch?

karol-szymanowski:
I just rebased it @Mokto

Mokto (approved):

chipzzz:
@karol-szymanowski , i think this is breaking using `existingSecretKey` worked fine , why was it removed? 
@Mokto

karol-szymanowski:
@chipzzz can you share content of the .value.yaml so I can replicate when it breaks?

chipzzz:
Disregard , it does not break.

Issue #1333, opened 2024-06-23
Bitnami version mismatch


benedikt-bartscher:
https://github.com/sentry-kubernetes/charts/pull/1301 introduced a regression.

With sourcemaps/memcached enabled like this:
```yaml
# sentry values.yaml
sourcemaps:
  # this enables memcached subchart
  enabled: true
```

Helm install fails with this error:

```log
rendering helm chart sentry for release sentry has failed: template: sentry/charts/nginx/templates/tls-secret.yaml:13:14: executing "sentry/charts/nginx/templates/tls-secret.yaml" at <include "common.labels.standard" (dict "customLabels" .Values.commonLabels "context" $)>: error calling include: template: sentry/charts/memcached/charts/common/templates/_labels.tpl:11:27: executing "common.labels.standard" at <include "common.names.name" .>: error calling include: template: sentry/charts/memcached/charts/common/templates/_names.tpl:11:18: executing "common.names.name" at <.Chart.Name>: nil pointer evaluating interface {}.Name
```

There are many issues like these in the bitnami helm chart repo. It happens if you mix and match bitnami charts with different versions.

Possible solution: Update memcached chart?

TartanLeGrand:
Why do you think it's related to nginx when the error you describe is a memcache error?

benedikt-bartscher:
take a close look at the error message. it's a version conflict between nginx and memcached

TartanLeGrand:
You are in the `23.9.1` ?

benedikt-bartscher:
This error occurs with every tagged version of this chart since #1301 was merged.

TartanLeGrand:
Okay I got the version: https://github.com/sentry-kubernetes/charts/releases/tag/sentry-v23.7.0

But your issue is linked with the issue:

- https://github.com/sentry-kubernetes/charts/issues/1315

Issue #1315, opened 2024-06-13
sentry-nginx   Readiness probe failed


MynameisFragile:
Nginx can not pass  Readiness probe and pod does not start with an error 
 Readiness probe failed: Get "http://10.8.32.55:8080/auth/login/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
 
 
 
 
 in  nginx pod logs i can endless auth tries: 
10.8.32.1 - - [13/Jun/2024:14:54:22 +0000] "GET / HTTP/1.1" 302 0 "-" "kube-probe/1.27" "-"
10.8.32.1 - - [13/Jun/2024:14:54:25 +0000] "GET /auth/login/ HTTP/1.1" 499 0 "http://10.8.32.55:8080/" "kube-probe/1.27" "-"

Never faced this issue before, but this is not the first time I deploy sentry
Why nginx can not auth and pass the readness prod, from where I should start? 

Best Regards

bmassemin:
Hi @MynameisFragile 

I'm trying to deploy this helm chart the last few days, and I observed the same issue:
- nginx has a healthcheck on the web pod, and not on itself
- 499 is a timeout since the web pod is not ready, therefore k8s doesn't route request to web pod
- the web pod is not ready because the database is not initialized yet
- the database is not initialized because helm waits for the deployment to succeed to start the db-init
- the deployment does not succeed because nginx and web are not ready.

my workaround to bootstrap a clean sentry cluster (with a new database):
- scaledown web and nginx replicas to 0, it will trigger helm hooks since the installation will be successful
- scale up web and nginx to 1 once the db-init task is completed.

I hope it'll help.

MynameisFragile:
@bmassemin Man , U saved my day, thx a lot , I owe u a beer)  I did not try u instruction , but I deployed the previous helm chart version and everything is working as expected, so look like las update broke something in a chart structure. Thank u a lot again , god bless u )

TartanLeGrand:
👀 : https://github.com/sentry-kubernetes/charts/pull/1301 & https://github.com/sentry-kubernetes/charts/pull/1313 

💡 : https://github.com/bitnami/charts/issues/26326

TartanLeGrand:
Okay, it's due to the new version of Nginx Bitnami chart, they have upgrade the readiness and the liveness on the chart.

Idea :bulb: 1:

- Stop use hooks for the migrations for make Sentry ready

Idea :bulb: 2:

[...]

Issue #1250, opened 2024-05-13
Relay issue after some time


leokster:
I created the ticket in the official get-sentry project but it seems to be related to the helm deployment:
https://github.com/getsentry/self-hosted/issues/3016

After few minutes (~30) the logging events for this project doesn't work anymore. Events still get sent to the envelope endpoint and I receive a event ID. All services are healthy and no single container crashes. The events just disappear. If I then create a new project and change the DSN everything works again and events get logged to the new project again for some time until it fails again.

I found a workaround by setting the relay mode to proxy but this is probably not the best thing to do.

Mokto:
This issue is stale because it has been open for 30 days with no activity.

TartanLeGrand:
Am on it, have the same issue

TartanLeGrand:
Hello :wave: @leokster,

Do you have this error with some Snuba troubles ? :thinking:

leokster:
@TartanLeGrand what do you mean?

TartanLeGrand:
In our case, we have the error on the relay but it's Snuba that has the problem.

xvilo:
Small sidenote for @leokster; setting the modr to proxy, instead of managed will mean it won't do datascrubbing anymore. Highly discouraged as you will get sensitive data logged such as user passwords. It took us months before we noticed this and changed back to managed mode. We changed some settings in the SDK to temp scrub sensitive fields there in the mean time

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #1343, opened 2024-06-29
feat(deps)!: sentry update to 24.7.1

TartanLeGrand:
This PR introduces updates to our Sentry Kubernetes configuration to align with the new features in Sentry 24.7.1. The latest version of Sentry autonomously handles Kafka migrations and topic management, prompting some significant changes in our setup.

#### Changes:
- 🗑 **Removed Kafka Topics Provisioning**: Since Sentry 24.6.0 manages Kafka topics internally, I have removed the Kafka topics previously defined in the `values.yaml`.
- 🔄 **Moved and Optimized Jobs**: Shifted several jobs to new locations and removed some redundant jobs. Considering the potential for further optimization, possibly through the use of init containers in future updates.
- ~~📂 **Reorganized Chart Directory**: Moved charts into a dedicated `charts` folder to streamline chart linting and testing processes.~~

#### Issues Addressed:
- This update resolves the following issues:
  - ~~**Issue #1315**~~
  - **PR #1379**
  - **Issue #1333**
  - Additionally, I'm think that the Kafka BC will resolve the relay issue we're seeing in **Issue #1250**.

#### Pending:
- 🐛 There's an outstanding issue with the onboarding page that seems to stem directly from Sentry. I'm in the process of investigating this bug.

#### Next Steps:
- ~~🧹 To maintain clarity and ease of review, I plan to split these changes into smaller, focused PRs before finalizing the merge.~~

#### Request for Feedback:
- 🔄 I welcome any feedback on the job optimizations and the proposed use of init containers for future improvements.

Looking forward to your thoughts and suggestions!


charts/sentry/templates/kafka-topics.job.yaml
adonskoy, line 1: A separate job is not needed to initialize topics. Just add the flag `--create-kafka-topics`  to the db-init job

charts/sentry/values.yaml
```diff
@@ -1797,111 +1797,6 @@ zookeeper:
...
-    # Note that snuba component might fail if you set `hooks.snubaInit.kafka.enabled` to `false` and remove the topics from this default topic list.
-    topics:
-      - name: events
-        config:
-          "message.timestamp.type": LogAppendTime
-      - name: event-replacements
-      - name: snuba-commit-log
-        config:
-          "cleanup.policy": "compact,delete"
-          "min.compaction.lag.ms": "3600000"
```
NikolaiIuzhaninYS: Could you please add these custom settings to the place where it should be now, or at least add topic customisation guide to the documentation?

TartanLeGrand: Hi @NikolaiIuzhaninYS, :wave: 

I'm not sure if this is possible. Sentry now automatically provisions Kafka topics and has also removed Kafka topic creation :
https://github.com/getsentry/self-hosted/commit/a6cb07691018bbe66360208b5f5f507f64b3f606

The cleanup policy can sometimes break the snuba api when you have too many events in Kafka.

Mokto: So it's not even possible to set partition counts ?

There used to be a env variable called `TOPIC_PARTITION_COUNTS`, I'm not sure it still works.


@TartanLeGrand what's your status on the PR ?

adonskoy: `TOPIC_PARTITION_COUNTS` for snuba conf only. I didn’t find a way to configure the number of partitions in the sentry code

I think we can accept this change as is and later implement an alternative way to create topics

dromadaire54: I've a question why the init db job are not a hook anymore @TartanLeGrand ?

TartanLeGrand: Because they launch only if all health rules are ok. The change in the last version of Bitnami Nginx broke this thing. For have the health rules ok on the Nginx we need to have migrate and init before. 😅

dromadaire54: It could be a good idea to add an issue to the nginx bitnami helmchart. I don't know if it's a normal behaviour.

TartanLeGrand: It's a normal behavior 😉

adonskoy: I think we then need an initСontainer in each sentry and snuba pod, which will check that the migrations have been applied and only then launch the primary container(s)

TartanLeGrand: Instead a job ? 🤔

adonskoy: Nope. initContainer's task is to check that all migrations have been applied by the job and only in this case launch the container with the new version of sentry. This will be an ideal replacement for helm hooks

TartanLeGrand:
LGTM, but IDK why the test does not work ...

Mokto:
I personally think it was better before when we could configure the topics.

Right now with the automated we enforce partition count = 1 everywhere?

TartanLeGrand:
> I personally think it was better before when we could configure the topics.
> 
> Right now with the automated we enforce partition count = 1 everywhere?

I thinks for the topics we need to make an PR on the official repo 😄 

I have revert my perf optimization and I let 3 replicas for the resources 👍

TartanLeGrand:
Let's make an inherited BC. I go refactor the PR for fix the Nginx before update. 😄

TartanLeGrand:
@Mokto add this to the same major release.

Mokto:
I'm still against merging such a PR. We should be able to set the partition counts.

adonskoy:
I suggest then to update only the version. 24.8.0 has already been released, and the chart still uses the May release..

Mokto:
yeah that makes sense.

TartanLeGrand:
24.8.0 doesn't work because we got an error of ClickHouse. But 24.7.1 does, I intend to migrate to 24.7.1 and fix 24.8.0 with the ClickHouse redesign. 🤔

Mokto:
@TartanLeGrand just so that we're clear, you still intend to remove your changes related to `--create-kafka-topics` ?

TartanLeGrand:
Hello @Mokto, yes am sure it's related to https://github.com/getsentry/self-hosted/pull/3121.

Let to sentry the responsibility of that.

Mokto:
As I said before, I don't want our users to lose the ability to set custom partition counts. 

The self-hosted project is different as everything runs on 1 machine.

Mokto:
Thanks! I think it's ready to be merged?

fseguin-pass:
FYI there is a security fix in 24.7.1 cf https://github.com/getsentry/sentry/security/advisories/GHSA-fm88-hc3v-3www

TartanLeGrand:
> Thanks! I think it's ready to be merged?

Nop, fail install with this version

TartanLeGrand:
@Mokto

I understand your concerns about managing Kafka partitions, but I believe we should let Sentry handle the topic creation as outlined in their [Kafka schemas repository](https://github.com/getsentry/sentry-kafka-schemas/tree/main/topics).

In my opinion, it's the software's responsibility not the infrastructure's to manage kafka topics provisioning. In a modern software approach, the infrastructure should provide the necessary resources for the software to function properly, while the software takes care of provisioning what it needs to operate efficiently.

Mokto:
Agreed but right now the software doesn't support it.

Mokto:
It's now using 24.7.1.

PR #1355, opened 2024-07-08
feat: add parameters for web workers TTL

shcherbak:


Mokto (approved):

x-504:
@Mokto What was the reason behind this change?.

shcherbak:
@albertollamaso
our team suffered from an error when long working web-deployment blocked user logins with message "session expired".
We investigated this bug but no luck, restart always helped. Less worker ttl is a workaround for this sentry behaviour

PR #1356, opened 2024-07-09
feat(clickhouse): Allow overriding Clickhouse host

nebulatgs:
When a cluster is IPv6 only, the Clickhouse health checks will fail due to Clickhouse binding to `0.0.0.0`.

This PR allows the Clickhouse `listen_host` to be overridden, so you can set it to `::` for IPv6 clusters.

### DIFF
--- a/charts/clickhouse/Chart.yaml
+++ b/charts/clickhouse/Chart.yaml
@@ -12,6 +12,6 @@ keywords:
 name: clickhouse
 sources:
   - https://github.com/sentry-kubernetes/charts
-version: 3.10.0
+version: 3.9.0
 maintainers:
   - name: sentry-kubernetes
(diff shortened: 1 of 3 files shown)


alexng353:
👍

I actually moved off of clickhouse because it couldn't bind to an ipv6, I'm so grateful to you, nebulatgs, for bringing this to light. If this gets merged into clickhouse I'll finally be able to click houses again. Thank you so much.

charts/clickhouse/Chart.yaml
```diff
@@ -12,6 +12,6 @@ keywords:
 name: clickhouse
 sources:
   - https://github.com/sentry-kubernetes/charts
-version: 3.9.0
+version: 3.10.0
```
Mokto: Could you remove that change please?

nebulatgs: Done! Thought that was needed to satisfy the CI.

Mokto:
The CI has just been fixed by @TartanLeGrand.

Could you please rebase your branch?

PR #1357, opened 2024-07-09
chore: release develop

Mokto:
:robot: I have created a release *beep* *boop*
---

<details><summary>sentry: 1.0.0</summary>

### Features

* add parameters for web workers TTL ([#1355](https://github.com/sentry-kubernetes/charts/issues/1355)) ([a1b218f](https://github.com/sentry-kubernetes/charts/commit/a1b218f69a8ea20a987e11a94dbf052d5a05d3a8))

### Bug Fixes

* remove 'profiling-global-suspect-functions' as it is not supported on self hosted ([#1358](https://github.com/sentry-kubernetes/charts/issues/1358)) ([25004f6](https://github.com/sentry-kubernetes/charts/commit/25004f67e4cba551bb78d5c42af80d2e631c50de))
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #1360, opened 2024-07-10
ci: fix incremental versionning

TartanLeGrand:
Fix: https://github.com/sentry-kubernetes/charts/pull/1357


Mokto (approved):
Thanks, that was really annoying!

PR #1369, opened 2024-07-24
feat: update config Geo

arturmon:
Added support for downloading Geoip database from maxmind.

### DIFF
--- a/charts/sentry/templates/pvc-geoip.yaml
+++ b/charts/sentry/templates/pvc-geoip.yaml
@@ -10,7 +10,7 @@ metadata:
     heritage: "{{ .Release.Service }}"
 spec:
   accessModes:
-    - ReadWriteOnce
+    - ReadWriteMany
   resources:
     requests:
       storage: {{ .Values.geodata.persistence.size }}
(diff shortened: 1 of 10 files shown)


charts/sentry/templates/pvc-geoip.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: data-sentry-geoip
+  labels:
+    app: sentry
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  accessModes:
+    - ReadWriteOnce
```
adonskoy: In this volume access mode, all pods must be on the same node. You can change it to RWM, but then you need to specify in the readme that the storage provider must support this mode

Or download the database every time and save it to an ephemeral volume via the init container

arturmon: @adonskoy  I made rwx with a note in the readme.  The EmptyDir storage will need to be initialized in each service of the container, and plus it still uses memory, which can, in theory, lead to an oom kill. Which is not very good.

sdernbach-ionos:
I would love to see this merged as I think it is much easier to setup geoip like this then creating all the storage stuff on your own 👍🏻

Mokto:
It has been implemented upstream.

PR #1376, opened 2024-08-01
chore: Set externalClickhouse singleNode to false as default

chipzzz:
Clickhouse is never deployed as a single node by default. This parameter is even set to false when using Clickhouse that's included with this chart.

This parameter controls how Snuba views the Clickhouse deployment.

Overlooking to set this parameter to false when using external Clickhouse causes missing events in Sentry.

### DIFF
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -1787,7 +1787,7 @@ externalClickhouse:
   ## (https://clickhouse.tech/docs/en/operations/server-configuration-parameters/settings/#server-settings-remote-servers)
   ## or by executing `select * from system.clusters`
   ##
-  # clusterName: test_shard_localhost
+  clusterName: default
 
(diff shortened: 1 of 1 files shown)


patsevanton:
Sentry don`t support Distributed table in Clickhouse
snuba.clickhouse.errors.ClickhouseWriterError: Method write is not supported by storage Distributed with more than one shard and no sharding key provided

seborys40:
Yes it does I'm running distributed tables across replicas and so is everyone else .

Sharding is harder but is also possible, problem revolves around migrations.

What you're referring to is old and can be worked around .

chipzzz:
This was recently changed to replica 1 vs it being 3,  so this solves the problem on the other side . - Closing
https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/values.yaml#L1788

PR #1377, opened 2024-08-01
fix(nginx): custom readiness probe

TartanLeGrand:
Fix the readiness probe :smile:


TartanLeGrand:
@Mokto validate this please ~~and we can go to repear the celery workers. For the next fix.~~

PR #1379, opened 2024-08-02
fix!: deployment default values

TartanLeGrand:
This PR introduces key fixes to the Sentry Helm chart, adjusting default values to improve resource usage and reliability. These changes include:

- **Liveness Probe Adjustments 🛠️**: Enabled livenessProbe for better stability.
- **Nginx Readiness Probe 🧪**: Added a custom TCP readiness probe for more accurate service checks.
- **Resource Optimization 📉**: Lowered replica counts for several components to streamline deployments.
- **Ingress Controller Enabled 🌐**: Defaulted to enabling the ingress controller.

These fixes will help ensure the chart functions correctly and that tests pass reliably.

### DIFF
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -197,7 +197,7 @@ sentry:
 
   worker:
     enabled: true
-    replicas: 3
+    replicas: 1
     # concurrency: 4
     env: []
     resources: {}
@@ -214,7 +214,7 @@ sentry:
       maxReplicas: 5
       targetCPUUtilizationPercentage: 50
     livenessProbe:
-      enabled: false
+      enabled: true
       periodSeconds: 60
       timeoutSeconds: 10
       failureThreshold: 3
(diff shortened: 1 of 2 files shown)


TartanLeGrand:
@Mokto 🍾

charts/sentry/values.yaml
```diff
@@ -1631,7 +1639,7 @@ nginx:
   #   name: nginx
 
 ingress:
-  enabled: false
+  enabled: true
```
adonskoy, line 1642: Why did you decide to turn on ingress by default?

TartanLeGrand: Thank you for your question! I decided to enable Ingress by default to streamline the deployment process for most users. Enabling Ingress out-of-the-box simplifies the configuration for environments where external access to the services is required, which is a common scenario.

Mokto (approved):
Nice thanks

charts/sentry/values.yaml
themightychris: @TartanLeGrand what's the value of enabling ingress though without pairing it with configuring a hostname? New users will need to configure a working hostname for enabling ingress to do anything, there's no way to make ingress work out-of-the-box with zero configuration, so flipping `enabled` to true might as well accompany configuring a hostname so that invalid manifests aren't produced by default

PR #1392, opened 2024-08-22
fix: revert ClickHouse replicas number

zacharyarnaise:
Revert the ClickHouse replicas value modified by this previous PR:
- https://github.com/sentry-kubernetes/charts/pull/1379

AFAIK you can't simply change the number of ClickHouse replicas on an existing installation and except it to work
As of today, in this chart setting *x* ClickHouse replicas will result in *x* shards created so the data is split across multiple servers.

I suggest we simply revert this for now.
I understand it would be great to provide better defaults for anyone doing a fresh install of this chart, so I think the best course of action would be to integrate the change I reverted in a major version release instead and inform on the consequences of this change.

### DIFF
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -1735,8 +1735,8 @@ config:
 
 clickhouse:
   enabled: true
-  replicas: "1"
   clickhouse:
+    replicas: "3"
     imageVersion: "21.8.13.6"
     configmap:
       remote_servers:


ivoistaken:
Absolutely.
You can't change replicas, this breaks sharding. Clickhouse does not automatically recalculate/redistribute shards.
Here is more info: https://clickhouse.com/docs/en/guides/sre/scaling-clusters

Mokto:
Thanks.

zacharyarnaise:
@Mokto Sorry but I think we might need to do the same thing with the replicas of `zookeeper-clickhouse` as well. I'm investigating [the failed check](https://github.com/sentry-kubernetes/charts/actions/runs/10504084881/job/29098773613) on my PR to be sure.
I have to admit you merged this faster than I anticipated 😅

zacharyarnaise:
@Mokto Looks like the failure was unrelated, the checks passed just fine in https://github.com/sentry-kubernetes/charts/pull/1393 so I think we're good to go!

Mokto:
Thanks for the heads up

mosesdd:
I think its the same for Kafka Controllers isnt it?

deadlybore:
Well it was a major version bump from 23.x.x to 24.x.x and it was described as a breaking change in the changelog

Mokto:
That's true. Maybe I should kept the values from 24.0.0.

It's up to everyone to update the replica values

mosesdd:
Yeah major version with breaking changes and WITHOUT upgrade instructions. I'm using this chart for years now and thinks like that don't happen the first time. 
We need to make sure which replicas are safe to scale down and which have to be updated manually before upgrading to 24.0.0 and put this in the README.

deadlybore:
Seems pretty clear to me that if you upgrade without reviewing the change it will break https://github.com/sentry-kubernetes/charts/blob/74a9838390d9320cfab57f7588f48586f8597737/charts/sentry/CHANGELOG.md#-breaking-changes

mosesdd:
But its missing here https://github.com/sentry-kubernetes/charts/blob/74a9838390d9320cfab57f7588f48586f8597737/README.md

Also its not clear what exactly to change. Which replicas are safe to scale down and which are not. As we see, Clickhouse is not save to scale down and Kafka is not even working scaled down after clean install: https://github.com/sentry-kubernetes/charts/issues/1394

Mokto:
I reverted the change again in 25.0.0.

And added it to the changelog, just in case. even though it was stated in the release notes

mosesdd:
@Mokto well, now clickhouse replica is in the wrong section again so it uses default value of clickhouse chart with 3 replicas...

Mokto:
What a mess. it's fixed, sorry about that.

chipzzz:
Just to confirm here @Mokto @mosesdd @zacharyarnaise  @ivoistaken scaling up replicas should be okay correct? Also on the comment of sharding, if we have a single shard per replica does this still break if scaling down?

PR #1424, opened 2024-09-05
feat: Add HPA resources with labels and annotations for some deployments

patsevanton:
This pull request introduces Horizontal Pod Autoscaler (HPA) resources for various components within the Sentry Helm chart. The HPA resources are configured with labels and annotations to ensure consistency with the corresponding Deployment resources. This change enhances the management and visibility of these resources within the Kubernetes cluster.

**Changes Include:**

**Added HPA resources for the following components:**

- sentry-ingest-monitors
- sentry-ingest-occurrences
- sentry-ingest-profiles
- sentry-ingest-replay-recordings
- sentry-billing-metrics-consumer
- sentry-generic-metrics-consumer
- sentry-metrics-consumer
- sentry-symbolicator-api

**Add label and annotations:**
- sentry-ingest-consumer-attachments
- sentry-ingest-consumer-events
- sentry-vroom

Impact:

This change ensures that all HPA resources are consistently labeled and annotated, making it easier to manage and monitor them within the Kubernetes environment. It also aligns the HPA resources with the existing Deployment resources, improving overall consistency and maintainability.

PR #1432, opened 2024-09-07
feat: remove autoscaling configuration from sentry ingest components

patsevanton:
#### Changes:

This pull request removes the autoscaling configuration from the Sentry Ingest components. The primary reason for this change is that the number of replicas needs to be kept equal to the number of partitions in Kafka, which is managed manually. 

#### Details:

- **Deployment Files**:
  - Removed conditions for setting the number of replicas if autoscaling is disabled.
  - Removed autoscaling settings from the Deployment templates.

- **values.yaml**:
  - Removed autoscaling settings for all Sentry Ingest components.

#### Related Pull Request:
- #1424

Please review and merge if everything looks good.

Thanks!

### DIFF
--- a/charts/sentry/templates/sentry/ingest/monitors/deployment-sentry-ingest-monitors.yaml
+++ b/charts/sentry/templates/sentry/ingest/monitors/deployment-sentry-ingest-monitors.yaml
@@ -24,9 +24,7 @@ spec:
       app: {{ template "sentry.fullname" . }}
       release: "{{ .Release.Name }}"
       role: ingest-monitors
-{{- if not .Values.sentry.ingestMonitors.autoscaling.enabled }}
   replicas: {{ .Values.sentry.ingestMonitors.replicas }}
-{{- end }}
   template:
     metadata:
       annotations:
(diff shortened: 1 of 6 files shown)


patsevanton:
What do you think about this pull request ?

Mokto:
Autoscaling wise, you need to make sure sure replicas < partitions right ? It doesn't have to be the exact same number.

patsevanton:
I remove autoscaling from values.yaml for application without "kind: HorizontalPodAutoscaler"

https://github.com/sentry-kubernetes/charts/tree/develop/charts/sentry/templates/sentry/ingest/monitors
https://github.com/sentry-kubernetes/charts/tree/develop/charts/sentry/templates/sentry/ingest/occurrences
https://github.com/sentry-kubernetes/charts/tree/develop/charts/sentry/templates/sentry/ingest/profiles
https://github.com/sentry-kubernetes/charts/tree/develop/charts/sentry/templates/sentry/ingest/replay-recordings
https://github.com/sentry-kubernetes/charts/tree/develop/charts/sentry/templates/sentry/metrics/billing


Code don`t work:
```
    autoscaling:
      enabled: false
      minReplicas: 1
      maxReplicas: 3
      targetCPUUtilizationPercentage: 50
```

maths22:
FWIW we are using the autoscaling on the ingest consumers in production and it has been working well (we have more partitions created than we expect to generally have consumers provisioned, so it appropriately adds/removes work).  The key indeed is just that you can't have *more* workers than partitions or some of them will have no work to do.

patsevanton:
@maths22 how does hpa work for you if this file is not in the help chat (https://github.com/sentry-kubernetes/charts/tree/develop/charts/sentry/templates/sentry/ingest/monitors ) ?

PR #1446, opened 2024-09-11
fix: invalid parameter in deployments

Voodoolay:
Fixing additional parameters for sentry deployments when every additional parameter will cause the 'invalid parameter' error:
`
spec:
    containers:
    - args:
      - run
      - consumer
      - ingest-generic-metrics
      - --consumer-group
      - generic-metrics-consumer
      - --healthcheck-file-path
      - /tmp/health.txt
      - --
      - --max-poll-interval-ms
      - "90000"
`


patsevanton:
@Mokto Is it possible to merge this pull request?

Mokto (approved):

PR #1494, opened 2024-09-30
Configuring kafka to use zookeeper uses only brokers

maitredede:
Hello,

When I use Kafka with zookeeper (disabling kraft), the `db-check` job fails because expected service name does not exists.

values.yaml :
```yaml
kafka:
  networkPolicy:
    enabled: false
  kraft:
    enabled: false
  zookeeper:
    enabled: true
    port: 2181
  controller:
    replicaCount: 0
  broker:
    replicaCount: 3
```

### DIFF
--- a/charts/sentry/templates/hooks/sentry-db-check.job.yaml
+++ b/charts/sentry/templates/hooks/sentry-db-check.job.yaml
@@ -113,28 +113,30 @@ spec:
             while [ $KAFKA_STATUS -eq 0 ]; do
               KAFKA_STATUS=1
               {{- if .Values.kafka.enabled }}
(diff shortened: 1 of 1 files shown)


Mokto (approved):

deadlybore:
This is going to break the default setup of people using kraft

maitredede:
> This is going to break the default setup of people using kraft

Hello, it should not. 

In the init script, it checks if there is a value for `.Values.kafka.zookeeper.enabled`. The kafka chart has built-in checks to avoid kraft and zookeeper mode both configured, or has inconsistent configuration (it does not allow zookeeper enabled if kraft mode is enabled). So you have to explicitely disable kraft and enable zookeeper, like the values I put in my first comment.

If I am wrong, I may fix my code accordingly :smiley:

PR #1514, opened 2024-10-07
Allow customizing Kafka configuration for Relay + Fix sentry-user-create job not creating the user

nadecancode:
This PR adds `relay.processing.additionalKafkaConfig` to relay kafka in order to make sentry-relay be able to use more kafka settings like SSL. This also includes a fix for `user-create` job which job was exiting gracefully without creating the user.

### DIFF
--- a/charts/sentry/templates/hooks/user-create.yaml
+++ b/charts/sentry/templates/hooks/user-create.yaml
@@ -71,11 +71,12 @@ spec:
       - name: user-create-job
         image: "{{ template "sentry.image" . }}"
         imagePullPolicy: {{ default "IfNotPresent" .Values.images.sentry.pullPolicy }}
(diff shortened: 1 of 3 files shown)


Mokto:
So now what exit code does it return if the user exists already?

nadecancode:
> So now what exit code does it return if the user exists already?

The problem wasn't whether the user exists already, it just seems like the original command w/ args wasn't calling Sentry properly that Sentry CLI complained & exited. I don't believe my diff changes whatever error code is being returned, it merely fixes a (seemly) formatting issue.

Sorry I no longer have the exact message but I had to correct the job manifest to the one in PR in order to properly create an user.

Mokto:
Ok thanks!

PR #1522, opened 2024-10-08
fix: correct argument order of consumers

YannikSc:
Heya, I noticed an error message after upgrading sentry in some of my consumers

```
Usage: sentry run consumer [OPTIONS] CONSUMER_NAME [CONSUMER_ARGS]...
Try 'sentry run consumer --help' for help.

Error: No such option: --max-batch-size
```

after checking the template for it, I found a commit that already took care of a similar issue in the past (https://github.com/sentry-kubernetes/charts/commit/2861efa7192b8d8bc02835ef4ade16a21b2729f1). 

However there were some places and arguments missing in the move, so here we got!


Mokto (approved):
Thanks!

YannikSc:
Well, that was very fast. Thank you!

PR #1516, opened 2024-10-07
feat: add geoip support to sentry deployment

patsevanton:
This pull request introduces support for GeoIP in the Sentry deployment. The changes include:

1. **New Configuration Options**:
   - Added new configuration options for GeoIP in the `values.yaml` file.
   - The new options include `accountID`, `licenseKey`, `editionIDs`, `persistence`, `volumeName`, `mountPath`, and `path`.

2. **New Templates**:
   - Added a new `deployment-geoip-job.yaml` template to handle the GeoIP database update job.
   - Added a new `pvc-geoip.yaml` template for the PersistentVolumeClaim (PVC) to store the GeoIP database.
   - Added a new `secret-geoip-env.yaml` template to manage the environment variables required for GeoIP configuration.

3. **Volume Mounts**:
   - Added volume mounts for the GeoIP database in the `deployment-relay.yaml`, `deployment-sentry-web.yaml`, and `deployment-sentry-worker-events.yaml` templates.

4. **Documentation**:
   - Updated the `README.md` file with an example of how to configure GeoIP in the Sentry deployment.

PR #1524, opened 2024-10-08
fix: correct storageClass handling for geodata persistence

patsevanton:
### Description
This pull request addresses an issue with the storageClass configuration for geodata persistence in the Sentry Helm chart. The previous implementation had redundant comments and incorrect logic for setting the storageClassName.

#### Changes Made:
1. **README.md**:
   - Removed redundant comments related to storageClass configuration for geodata persistence.
   - Updated the default value for storageClass to an empty string.

2. **pvc-geoip.yaml**:
   - Corrected the logic for setting the storageClassName. Now, it only sets the storageClassName if it is defined in the values.yaml.

3. **values.yaml**:
   - Removed unnecessary comments related to storageClass configuration for geodata persistence.
   - Updated the default value for storageClass to an empty string.

Related Pull Requests:
- #1516

---
Feel free to customize the description to better fit your specific changes and context.

### DIFF
--- a/charts/sentry/README.md
+++ b/charts/sentry/README.md
@@ -1207,14 +1207,11 @@ geodata:
   licenseKey: "example"
   editionIDs: "example"
   persistence:
-    ## database data Persistent Volume Storage Class
     ## If defined, storageClassName: <storageClass>
-    ## If set to "-", storageClassName: "", which disables dynamic provisioning
     ## If undefined (the default) or set to null, no storageClassName spec is
     ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
     ##   GKE, AWS & OpenStack)
-    ##
-    # storageClass: "-"
+    # storageClass: ""
     size: 1Gi
   volumeName: "data-sentry-geoip"
   # mountPath of the volume containing the database
--- a/charts/sentry/templates/pvc-geoip.yaml
+++ b/charts/sentry/templates/pvc-geoip.yaml
@@ -14,9 +14,7 @@ spec:
   resources:
     requests:
       storage: {{ .Values.geodata.persistence.size }}
-{{- if (eq "-" .Values.geodata.persistence.storageClass) }}
-  storageClassName: ""
-{{- else }}
-  storageClassName: "{{ .Values.geodata.persistence.storageClass }}"
-  {{- end }}
+{{- if .Values.geodata.persistence.storageClass }}
+  storageClassName: {{ .Values.geodata.persistence.storageClass | quote }}
+{{- end }}
 {{- end }}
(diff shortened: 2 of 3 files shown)


patsevanton:
@sdernbach-ionos please review PR https://github.com/sentry-kubernetes/charts/pull/1524

sdernbach-ionos (approved):

sdernbach-ionos:
@patsevanton that was fast! Thank you! Somehow for us the installation of a fresh install is not working as the pvc does not get provisioned before the geo ip job runs (which depends on the pvc), does that work for you?

patsevanton:
@Mokto please review PR
@sdernbach-ionos I don not have account. I can not recheck. I will see what I can do.

patsevanton:
@sdernbach-ionos I try ReadOnlyMany and ReadWriteMany https://github.com/sentry-kubernetes/charts/pull/1526 but dont work in my cloud 

```
WaitForFirstConsumer  108s (x26 over 8m2s)  persistentvolume-controller  waiting for first consumer to be created before binding
```

sdernbach-ionos:
@patsevanton I answered in the PR to move the discussion there, thanks for trying!

patsevanton:
@sdernbach-ionos are we rollback an pull request https://github.com/sentry-kubernetes/charts/pull/1516 ?

sdernbach-ionos:
@patsevanton saw you closed it, we can also discuss here if you mean that? 
Do you think the support of ReadWriteMany is same as ReadOnlyMany? Does it make sense to have this configurable? As even ReadWriteOnce would work in case the pods that are reading are on the same node. But I think the pre-install definitely needs to be changed to post-install as the PVC needs to be created first.

patsevanton:
Few cloud providers support ReadWriteMany/ReadOnlyMany out of the box. So far, there is no point. Maybe in the future.

sdernbach-ionos:
@patsevanton if I got the access modes the right way the normal case is ReadWriteOnce which would work only if the pods are on the same node. If this is not the case the cloud would need to support ReadWriteMany or ReadOnlyMany. On the other side, the manual setup that is described, how does this work? You also have to setup a PVC and you have the same problems then, right? If the pods are on multiple nodes and your cloud only supports ReadWriteOnce you would need to create multiple PVCs with the file so web and relay could access it? 

Just saw that you edit your comment above and meant rolling back the PR, I'm personally in favor of keeping it (just fixing the pre-install > post-install) and if ReadWriteMany/ReadOnlyMany is the only option making it configurable or leaving it with ReadWriteMany (or making it configurable with a default) and naming this in the documentation (as it is already named that your cloud needs to support ReadWriteMany). For sure I can also go ahead and open a PR for this change.

patsevanton:
OK, I'll make a new PR based on https://github.com/sentry-kubernetes/charts/pull/1526 later.

PR #1526, opened 2024-10-09
ReadOnlyMany dont work in my cloud

patsevanton:


sdernbach-ionos:
@patsevanton I will just answer here instead of the old PR. I guess ReadOnlyMany will not work as the job has to write the database file to the storage?

patsevanton:
ReadOnlyMany – the volume can be mounted read-only by many nodes
By this method, multiple pods running on multiple nodes can use a single volume and read data.
If a pod mounts a volume with ReadOnlyMany access mode, other pod can mount it and perform only read operation. GCP supports this from GKE version [1.22](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/readonlymany-disks#requirements).
This means a volume can be mounted on one or many nodes of your Kubernetes cluster and you can only perform read operation.
You have one pod is running on node and you are reading stored file from volume. While on same volume you can not perform writes.
As it's ReadOnlyMany, if your pod is scheduled to another node, then also volume and data will be available to perform read operation.


But ReadOnlyMany does not work in all clouds

sdernbach-ionos:
@patsevanton got it, thanks! Do you think the support of ReadWriteMany is same as ReadOnlyMany ? Does it make sense to have this configurable? As even ReadWriteOnce would work in case the pods that are reading are on the same node

patsevanton:
Few cloud providers support ReadWriteMany/ReadOnlyMany out of the box. So far, there is no point. Maybe in the future.

PR #1530, opened 2024-10-12
fix: symbolicator storage class

xpicio:
Hello, I made a small fix to the symbolicator's `storageClassName`. The condition was previously reading the `storageClass` key, but the property was being set using `storageClassName`. 👋


patsevanton:
Hi!
Please dont change `charts/sentry/Chart.yaml`

xpicio:
Change removed

PR #1492, opened 2024-09-29
feat: reintroduce "Extend Redis functionality"

nadecancode:
This PR reintroduces https://github.com/sentry-kubernetes/charts/pull/914 with conflicts resolved.

Core Diff:
- Addressed https://github.com/sentry-kubernetes/charts/pull/914#discussion_r1391345725 with a workaround that declares an additional environment variable that references the externalRedis secret then use https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/ to plug Redis URL in. The dependency resolution is handled by Kubernetes so it is secure.
- Rename the password reference in values.yml for Redis subchart to `redis.auth.password` etc., this is because subchart relies on that path to create authentication so setting `redis.password` only will only *break* the integrated Redis connection (because Redis subchart will believe there is no password defined).
- Remove the need to specify brokerUrl secrets, it'll compute based on the existing secrets specified in the values
- Fixed misc issues that were failing the E2E

PR #1548, opened 2024-10-15
fix: external redis functionality for relay

iamyogeshg:
This PR is to extend the fix by #1492 

As per recommendation [here](https://github.com/getsentry/relay/issues/1722) we need to provide username:password format for redis uri. It appears that in the PR  #1492 the colon is missing for relay while it was correctly given in helper.tpl[here](https://github.com/sentry-kubernetes/charts/commit/0b7a7b4c874bf4d4a460c88bb259cab0e025f7ee#diff-1327e24804f05c729b65b01af5f5a8cd1c8496a44ed336e3339c20c4fd805dbaR652-R660)

We are using external redis from aws elasticache in non cluster mode and using chart version 25.11.1 with following config:

```
externalRedis:
  host: master.uat-XXXXXX.amazonaws.com
  existingSecret: sentry
  existingSecretKey: redis-password
  ssl: true

redis:
  enabled: false
``` 

The relay component is throwing errors:

```
2024-10-14T04:00:01.703356Z ERROR relay_server::services::processor: error processing envelope tags.project_key=834fe1715ec71e6042077d96dbde5ca4 error=failed to apply quotas error.sources=[failed to communicate with redis, failed to communicate with redis, NOAUTH: Authentication required.]
```

### DIFF
--- a/charts/sentry/templates/relay/deployment-relay.yaml
+++ b/charts/sentry/templates/relay/deployment-relay.yaml
@@ -99,7 +99,7 @@ spec:
                   name: {{ .Values.externalRedis.existingSecret }}
                   key: {{ default "redis-password" .Values.externalRedis.existingSecretKey }}
             - name: RELAY_REDIS_URL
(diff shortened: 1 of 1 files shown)


iamyogeshg:
@Mokto Please let me know if there is anything else I need to update. Thanks!

Mokto:
Thanks!

PR #1558, opened 2024-10-17
feat: add externalrabbitmq support

2005wind:
hi, i want to  add  externalrabbitmq support, pls review . tks .

### DIFF
--- a/charts/sentry/templates/sentry/_helper-sentry.tpl
+++ b/charts/sentry/templates/sentry/_helper-sentry.tpl
@@ -195,7 +195,7 @@ sentry.conf.py: |-
   # information on configuring your queue broker and workers. Sentry relies
   # on a Python framework called Celery to manage queues.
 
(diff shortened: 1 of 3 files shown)


nadecancode:
I believe externalRabbitmq is already supported, you just have to set rabbitmq.enabled to false AND specify a host.

charts/sentry/templates/sentry/_helper-sentry.tpl
```diff
@@ -195,8 +195,10 @@ sentry.conf.py: |-
   # information on configuring your queue broker and workers. Sentry relies
   # on a Python framework called Celery to manage queues.
 
-  {{- if or (.Values.rabbitmq.enabled) (.Values.rabbitmq.host) }}
```
nadecancode, line 198: This is not backwards compatible, my deployment relies on this to get the external rabbit running

nadecancode (changes requested):

Mokto:
External rabbitmq is already supported indeed.

Thanks for the PR anyways.

PR #1622, opened 2024-12-05
fix: missing flags required for Queues and Queries instrumentation

mzglinski:
Those 2 flags are in the official compose deployment: https://github.com/getsentry/self-hosted/blob/master/sentry/sentry.conf.example.py#L263

Only after enabling them, data started showing in Queues instrumentation and the query details in Queries. 

image


Mokto (approved):
Thanks!

PR #1626, opened 2024-12-08
feat: enhance worker deployments with additional packages and CA certificates

patsevanton:
This pull request introduces several enhancements to the Sentry worker deployments:

1. **Additional Package Installation**: 
   - Workers now have the capability to install additional Python packages during startup. This is controlled by the `installAdditionalPackages` flag in the `values.yaml` file.
   - The following packages are installed if the flag is disabled:
     - `django-multidb-router`
     - `sentry-nodestore-elastic`
     - `sentry-s3-nodestore` (from a specific release)

2. **CA Certificates Management**:
   - Workers can now manage custom CA certificates by mounting a secret containing the certificates to `/usr/local/share/ca-certificates/`.
   - The certificates are then copied to the appropriate location and updated using `update-ca-certificates`.
   - This feature is controlled by the `caCertificatesSecret` field in the `values.yaml` file.

### Changes

- **Deployment Templates**:
  - Modified the `command` and `args` sections in the worker deployment templates to include a bash script that handles the installation of additional packages and CA certificates.
  - Added volume mounts and volume definitions for the CA certificates secret.

- **Values Configuration**:
  - Introduced new configuration options in the `values.yaml` file:
    - `sentry.worker.installAdditionalPackages`: Boolean flag to enable/disable the installation of additional packages.
    - `sentry.worker.caCertificatesSecret`: Name of the secret containing custom CA certificates.

### Check pip packages

```
pip list | grep -E 'django|nodestore'
django-crispy-forms                1.14.0
django_csp                         3.8
django-multidb-router              0.10
django-pg-zero-downtime-migrations 0.13
djangorestframework                3.15.2
sentry-nodestore-elastic           1.0.1
sentry-s3-nodestore                1.0.3
```

### Check certificates
```
import os
import ssl
import certifi
import re
from cryptography import x509
from cryptography.hazmat.backends import default_backend

def list_certificates():
    # Get the path to the cacert.pem file
    cacert_path = certifi.where()

    # Open the file containing the certificates
    with open(cacert_path, 'r') as f:
        pem_data = f.read()

    # Use a regular expression to extract certificates
    certificate_pattern = re.compile(r'-----BEGIN CERTIFICATE-----.+?-----END CERTIFICATE-----', re.DOTALL)
    certificates = certificate_pattern.findall(pem_data)

    # Display information about each certificate
    for i, cert_pem in enumerate(certificates):
        try:
            # Load and parse the certificate
            cert = x509.load_pem_x509_certificate(cert_pem.encode(), default_backend())
            subject = cert.subject
            issuer = cert.issuer
            not_valid_before = cert.not_valid_before
            not_valid_after = cert.not_valid_after

            # Print certificate information
            print(f"Certificate #{i + 1}:")
            print(f"  Subject: {subject}")
            print(f"  Issuer: {issuer}")
            print(f"  Valid from: {not_valid_before}")
            print(f"  Valid until: {not_valid_after}")
            print()
        except Exception as e:
            # Print error message if there's an issue with the certificate
            print(f"Error processing certificate #{i + 1}: {e}")

# Example usage
list_certificates()
```

### DIFF
--- a/charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
+++ b/charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
@@ -79,8 +79,8 @@ spec:
         command: ["/bin/bash", "-c"]
         args:
           - |
-            {{- if .Values.sentry.workerEvents.installAdditionalPackages }}
-            pip install django-multidb-router sentry-nodestore-elastic https://github.com/pavels/sentry-s3-nodestore/releases/download/v1.0.3/sentry-s3-nodestore-1.0.3.tar.gz
+            {{- if .Values.sentry.worker.installAdditionalPackages }}
+            pip install {{ range .Values.sentry.worker.installAdditionalPackages }}{{ . }} {{ end }}
             {{- end }}
             # Copy certificates from secret to /usr/local/share/ca-certificates/
             mkdir -p /usr/local/share/ca-certificates/
--- a/charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
+++ b/charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
@@ -76,10 +76,20 @@ spec:
       - name: {{ .Chart.Name }}-worker
         image: "{{ template "sentry.image" . }}"
         imagePullPolicy: {{ default "IfNotPresent" .Values.images.sentry.pullPolicy }}
-        command: ["sentry"]
+        command: ["/bin/bash", "-c"]
         args:
-          - "run"
-          - "worker"
+          - |
+            {{- if .Values.sentry.workerEvents.installAdditionalPackages }}
+            pip install django-multidb-router sentry-nodestore-elastic https://github.com/pavels/sentry-s3-nodestore/releases/download/v1.0.3/sentry-s3-nodestore-1.0.3.tar.gz
+            {{- end }}
+            # Copy certificates from secret to /usr/local/share/ca-certificates/
+            mkdir -p /usr/local/share/ca-certificates/
+            cp /etc/ssl/certs/* /usr/local/share/ca-certificates/
+            for c in $(ls -1 /usr/local/share/ca-certificates/); do
+                cat /usr/local/share/ca-certificates/$c >> $(python3 -m certifi) && echo >> $(python3 -m certifi)
+            done
+            update-ca-certificates
+            sentry run worker
           - "-Q"
           - {{ .Values.sentry.workerEvents.queues }}
           {{- if .Values.sentry.workerEvents.concurrency }}
@@ -122,6 +132,11 @@ spec:
 {{- if .Values.sentry.workerEvents.volumeMounts }}
 {{ toYaml .Values.sentry.workerEvents.volumeMounts | indent 8 }}
 {{- end }}
+        {{- if .Values.sentry.worker.caCertificatesSecret }}
+        - name: ca-certificates
+          mountPath: /usr/local/share/ca-certificates
+          readOnly: true
+        {{- end }}
   {{- if .Values.sentry.workerEvents.livenessProbe.enabled }}
         livenessProbe:
           periodSeconds: {{ .Values.sentry.workerEvents.livenessProbe.periodSeconds }}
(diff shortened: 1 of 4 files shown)


charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
```diff
@@ -76,10 +76,18 @@ spec:
       - name: {{ .Chart.Name }}-worker
         image: "{{ template "sentry.image" . }}"
         imagePullPolicy: {{ default "IfNotPresent" .Values.images.sentry.pullPolicy }}
-        command: ["sentry"]
+        command: ["/bin/bash", "-c"]
```
adonskoy, line 79: It's better to use init container and emptyDir. Нou will get the same result without changing the main container command

charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
```diff
@@ -76,10 +76,18 @@ spec:
...
         args:
-          - "run"
-          - "worker"
+          - |
+            {{- if .Values.sentry.worker.installAdditionalPackages }}
```
adonskoy, line 82: Installing packages at runtime is not the best practice and will take much longer at startup compared to a custom image. Is this really necessary?

charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
```diff
@@ -76,10 +76,18 @@ spec:
...
+            pip install {{ range .Values.sentry.worker.installAdditionalPackages }}{{ . }} {{ end }}
+            {{- end }}
+            mkdir -p /usr/local/share/ca-certificates/
+            for c in $(ls -1 /usr/local/share/ca-certificates/); do
```
adonskoy, line 86: How about adding this to the template and adding it to all sentry workloads?

After all, CAs may be needed throughout the entire system, right?

adonskoy (changes requested):

charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
patsevanton: Yes, I don't see any other options

charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
patsevanton: Can you write the example code?

charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
patsevanton: What commands should I run in the init container?

adonskoy: Same commands, but in different container and mount emptyDir for `/usr/local/share/ca-certificates/`. Like [init container in relay](https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/templates/relay/deployment-relay.yaml#L82)

patsevanton: @adonskoy could you create MR with you idea?

patsevanton: @adonskoy there is no point in adding certificates to python storage using init container since python certificate storage in the main container is dynamic and can change from version to version

adonskoy: Have you tried using the `REQUESTS_CA_BUNDLE` env? Should work for sentry

patsevanton: I also need to install additional python packages. They cannot be installed via init container

charts/sentry/templates/sentry/worker/deployment-sentry-worker-events.yaml
adonskoy: It seems to me that this can be solved in a more universal way: add the ability to set a custom command to launch the container. Then you can add what you need in your case

patsevanton: Can you create new MR with you idea?

patsevanton: @adonskoy Can you create new MR with you idea?

tuandang-diag:
Can someone continue this PR? @patsevanton @adonskoy

Issue #1383, opened 2024-08-09
Merge THIS PR #915 - FIXES incorrect default config for internal_replication


chipzzz:
Merge this PR please, the current config is entirely incorrect and [default value in the helm chart are NOT valid or true](https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/values.yaml#L1734).

https://github.com/sentry-kubernetes/charts/pull/915

Today the helm chart still shows true. So anyone coming into chart is thinking this is set true by default. The deployment itself is incorrect as the parameter is deployed on the wrong section.

This could be drastic when trying to fully fix distributed/sharded mode of snuba/clickhouse. Based on the findings mentioned in the official CH docs the placement of this config is entirely incorrect so this has been always false as mentioned by @zetaab .

This PR should be merged so that there's no confusion about default config. Today the config is still deployed in the wrong spot making the setting False,[ yet the helm chart says it's true](https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/values.yaml#L1734)

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #1633, opened 2024-12-14
fix(clickhouse-chart): correct placement of internal_replication

pyhp2017:
This PR addresses a configuration issue in the ClickHouse chart where the internal_replication setting was incorrectly placed inside the "replica" tag. The correct placement for internal_replication is inside the "shard" tag, as per the expected configuration schema.

### DIFF
--- a/charts/clickhouse/templates/configmap-config.yaml
+++ b/charts/clickhouse/templates/configmap-config.yaml
@@ -38,8 +38,8 @@ data:
             <{{ include "clickhouse.fullname" . }}>
             {{- range untilStep 0 (int .Values.clickhouse.replicas) 1 }}
                 <shard>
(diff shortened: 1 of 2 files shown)


chipzzz:
ref: https://github.com/sentry-kubernetes/charts/issues/1383
https://github.com/sentry-kubernetes/charts/pull/915

chipzzz:
This is still not approved/merged  

@Mokto / @patsevanton - could we please finally fix this as the current config is invalid.

patsevanton:
@Mokto Approve 
Prof https://chistadata.com/knowledge-base/cluster/
https://dba.stackexchange.com/questions/292523/clickhouse-replication-without-sharding
https://dev.to/zergon321/creating-a-clickhouse-cluster-part-ii-replication-23mc

PR #1634, opened 2024-12-15
Sampled default rate

artBass-rip:


patsevanton:
@artBass-rip please create new MR with commit "feat: sampled default rate"

Mokto:
It's fine I can change that when I merge.

Issue #1681, opened 2025-01-26
Symbolicator docker image tag not found


bmassemin:
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/sentry-kubernetes/charts/issues)

### Describe the bug (actual behavior)

`failed to resolve reference "docker.io/getsentry/symbolicator:24.12.2": docker.io/getsentry/symbolicator:24.12.2: not found`

Tag not found on docker hub: https://hub.docker.com/r/getsentry/symbolicator/tags?name=24.12.2

That's weird because `24.12.1` is still available

### values.yaml

```
...
    symbolicator:
      enabled: true
...
```

### Helm chart version

26.12.0

### Steps to reproduce

Install the chart on a fresh node without docker cache

patsevanton:
Indeed, symbolicator does not have an image with version 24.12.2. What are the suggestions? we can't make a new sentry kubernetes tag 24.12.1

WallferJDI:
Override tag worked for me
```
images:
  symbolicator:
    repository: getsentry/symbolicator
    tag: 24.12.1
```

Mokto:
I'm thinking of https://github.com/sentry-kubernetes/charts/pull/1685

patsevanton:
@bmassemin please check new version

bmassemin:
Thanks 🙏

PR #1683, opened 2025-01-28
docs: Symbolicator docker image tag 24.12.2 not found

patsevanton:
Info in issue https://github.com/sentry-kubernetes/charts/issues/1681

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ helm install my-sentry sentry/sentry --wait --timeout=1000s
 
 For now the full list of values is not documented, but you can get inspired by the `values.yaml` specific to each directory.
 
+## Upgrading from 26.11.2 Version of This Chart to 26.12.0
+
+Symbolicator docker image tag 24.12.2 not found
+
 ## Upgrading from 25.x.x Version of This Chart to 26.x.x
 
 Make sure to upgrade to chart version 25.20.0 (Sentry 24.8.0) before upgrading to 26.x.x.


Mokto:
Shouldn't we just rollback the version then ?

ivoistaken:
https://github.com/getsentry/symbolicator/releases
There is no 24.12.2 (tags/releases), why not just use 24.12.1 ?

patsevanton:
@Mokto do you mean delete the tag? 
@ivoistaken do you suggest writing to use tag 24.12.1 for symbolicator?

ivoistaken:
> @Mokto do you mean delete the tag? @ivoistaken do you suggest writing to use tag 24.12.1 for symbolicator?
Yes, just try using the latest 24.12 (in this case 24.12.1)

Mokto:
I suggest we change the version there:
https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/Chart.yaml#L6

patsevanton:
> I suggest we change the version there: https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/Chart.yaml#L6

to do this, you must first delete the tag.

Mokto:
I don't see why. We mark 26.12.1 as a fix deploying version 24.12.1

patsevanton:
we currently have the app version 24.12.2, so you're suggesting making a new tag with the app version 24.12.1. do you think it's a good idea to downgrade versions?

Mokto:
I think it's fine as long as we only downgrade patch version.

I would be worried about minor versions.

patsevanton:
@Mokto OK. the only question is how to do it right.

PR #1693, opened 2025-02-01
docs:  Added existingSecret comment in Sentry values.yaml

Xantibody:
### Background
When setting up Sentry configuration in values.yaml, I encountered difficulties due to the lack of concrete examples. Since it was unclear whether the configuration was implemented as expected, I had to dive into the source code to understand the correct configuration patterns.

### Proposed Changes
- Added comments and examples for existingSecret configuration in Sentry's values.yaml

### DIFF
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -1828,36 +1828,36 @@ google: {}
 #   clientId: ""
 #   clientSecret: ""
 #   existingSecret: ""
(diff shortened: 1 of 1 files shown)


patsevanton:
Fix Conventional Commits Chec https://github.com/sentry-kubernetes/charts/pull/1697

Mokto:
Maybe we should remove that check before when i squash & merge I make sure the commit is conventional anyways

PR #1712, opened 2025-02-24
chore: cleanup unused variables

mikmatko:
`sentry.subscriptionConsumerSessions`, `snuba.sessionsConsumer` and `snuba.subscriptionConsumerSessions` are not referenced anywhere in templates. I assume those are some old services which were replaced by something else and was forgotten to cleanup values.yaml and README.


Mokto:
Thanks!

PR #1756, opened 2025-05-10
fix(clickhouse): add missing configOverride in statefulset-clickhouse-replica

pyhp2017:
The configOverride section was missing in the ClickHouse statefulset for replicas.


pyhp2017:
Hello @Mokto,
accidentally i closed the previous PR, here is the new one without conflict

Mokto:
Thanks!

PR #1789, opened 2025-06-16
fix: remove reference to secret-key from NOTES.txt

poja-soptim:
`system.secret-key` has been removed in chart version 16


Mokto:
Thanks

PR #1810, opened 2025-07-11
feat: Add errors-only mode from original self-hosted compose

### LINKED ISSUE #1372: Errors-only deployment mode?
Hi there,

I see that Sentry Self-hosted has an new beta feature called "Errors-only":
https://develop.sentry.dev/self-hosted/#errors-only-self-hosted-sentry

>Starting from 24.8.0+, users will have the ability to choose between two distinct types of self-hosted Sentry deployments: errors only or feature complete..
>
>Errors Only self-hosted offers a small subset of features with an emphasis on minimizing system resources. This lightweight option includes only the following features:
>
>* Issues
>* Alerts
>* Integrations
>* Dashboards
>* Releases
>* Discover
>
>In order to enable errors only self-hosted, you'll need to update your [.env file](https://github.com/getsentry/self-hosted/blob/master/.env) to include COMPOSE_PROFILES=errors-only.

I assume such a deployment would be simpler and might need fewer components deploying? 

Is this something that is on the roadmap to be supported by this Helm chart? Thanks!

motoki317:
This pull request adds "errors-only" mode as first addressed in #1372.

I've looked up [the original docker compose](https://github.com/getsentry/self-hosted/blob/master/docker-compose.yml) file, and wrapped components which had `profiles: [feature-complete]` with if statements that check for active profiles.
While adding if statements, I've noticed some components are missing in this chart that are present in the original docker-compose.yml file - I've not addressed these in this PR.

This mimics the original docker compose behavior, so this should be easy to maintain.
However, I also noticed that `sentry.features` values is also present in the values.yaml file which seems to control the ConfigMap contents but not enable/disable the deployments themselves.
Perhaps a future task would be to "merge" these values toggle functions, so the chart users can choose which features to enable and deploy only the minimal components.

However, for now I've stuck with a simple approach of mimicking the compose profiles behavior for easier maintenance.
Please feel free to add corrections in this branch or discuss how we should implement "errors-only" mode in this chart.

Thanks!

closes #1372 (although already closed as stale)

### DIFF
--- a/charts/sentry/templates/sentry/ingest/feedback/deployment-sentry-ingest-feedback.yaml
+++ b/charts/sentry/templates/sentry/ingest/feedback/deployment-sentry-ingest-feedback.yaml
@@ -1,3 +1,4 @@
+{{- if has "feature-complete" .Values.profiles }}
 {{- if .Values.sentry.ingestFeedback.enabled }}
 apiVersion: apps/v1
 kind: Deployment
@@ -177,4 +178,5 @@ spec:
       {{- if .Values.sentry.ingestFeedback.priorityClassName }}
       priorityClassName: "{{ .Values.sentry.ingestFeedback.priorityClassName }}"
       {{- end }}
+{{- end }}
 {{- end }}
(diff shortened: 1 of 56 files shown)


Mokto (approved):

Mokto:
Thanks!

PR #1817, opened 2025-07-21
feat: Support `containerSecurityContext` for Symbolicator Cleaner container

Charismara:
Providing `securityContext` for all containers is required in some environments.


Mokto:
thanks

Issue #1828, opened 2025-07-29
Bitnami chart and container dependencies


lasermoth:
Bitnami has recently [announced changes to their catalog](https://github.com/bitnami/charts/issues/35164), which includes both charts and their images. 

Images will be moved to `docker.io/bitnamilegacy` but will receive no future updates.

What will be the plan for maintaining this chart going forward given the changes to the upstream Bitnami dependencies defined in this chart?

lasermoth:
Linking to this discussion which also raises it https://github.com/sentry-kubernetes/charts/discussions/1825

mikmatko:
@Mokto Note that this will break all new installs of this chart starting **August 28th, 2025**, as well as many existing installations (if pods are scheduled to a node where the image hasn't been pulled in).

As a short term bandage, all bitnami repository references could be overridden in [values.yaml](https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/values.yaml). Currently they are not specified in values.yaml at all, but rather take default values from their own charts, which will have incorrect values beginning from August 28th.

Mokto:
Crazy. So we just need to move things to docker.io/bitnamilegacy for now to save the situation, right?

mikmatko:
> Crazy. So we just need to move things to docker.io/bitnamilegacy for now to save the situation, right?

Yes, exactly.

TartanLeGrand:
@Mokto You are right, but we now have two considerations:

1. In a closed environment, such as a private network, some CVEs might be "acceptable."
2. In a world-facing environment, CVEs are not acceptable.

We already have a lot of updates to make (Redis, NGINX, RabbitMQ, etc.). Let's apply those updates to the latest versions before switching to the Bitnami legacy repository. We can then decide later about the future of the Bitnami-based chart.

fildafer:
I'm one of the users who will be impacted by Bitnami's new policy. Are you folks planning to release a new version of the Chart with the updated dependencies or we should fix it locally? Thanks.

TartanLeGrand:
I think we can quickly fix the container image repository to legacy Bitnami Repo.

gecube:
I would be happy to see the support of Strimzi Kafka. It is the right way, I believe. Should I open a new issue?

Mokto:
This issue is stale because it has been open for 30 days with no activity.

gecube:
the issue should be resolved. The easiest way - to override images to point to bitnamylegacy repo on docker hub.

[...]

PR #1840, opened 2025-08-11
fix(deps): Use bitnami images from docker.io/bitnamilegacy 

mikmatko:
A quick bandage for #1828. Some users may have such configuration that other updates are also needed, but this should work for the default installation.

### DIFF
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -2120,6 +2120,8 @@ openai: {}
 
 nginx:
   enabled: true  # true, if Safari compatibility is needed
+  image:
+    repository: bitnamilegacy/nginx
   containerPort: 8080
   existingServerBlockConfigmap: '{{ template "sentry.fullname" . }}'
   resources: {}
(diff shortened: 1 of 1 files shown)


rwojsznis:
FYI I ran into this argocd error when trying to explicitly specify image version in their redis chart:

> (...) Original containers have been substituted for unrecognized ones. Deploying this chart with non-standard containers is likely to cause degraded security and performance, broken chart features, and missing environment variables.%0A%0AUnrecognized images:%0A  (...) global parameter 'global.security.allowInsecureImages' to true.%0AFurther information can be obtained

Not sure if this check is done in all their charts - the redis one I'm using is pretty fresh (v22)

mikmatko:
Ping @Mokto, I think you'll want to release a new version with this included :)

Mokto (approved):

Mokto:
Could you rebase please?

TartanLeGrand (approved):
LGTM!

Mokto:
thanks a lot!

manishrawat1992:
Free software is degrading fast

PR #1851, opened 2025-08-28
feat: upgrade to 25.8.0

mzglinski:
This update includes:

*   Sentry updated to v25.8.0
*   Removal of intermediate containers for EAP features
*   Uptime monitoring (`features.enableUptime`)
*   Cleanup of feature flags: Removed obsolete flags and realigned chart set with [self-hosted](https://github.com/getsentry/self-hosted)
*   Move to ghcr.io images (Sentry no longer publishes to Docker Hub)

PR #1853, opened 2025-08-28
feat: persistence improvements

mzglinski:
I just wanted to enable geoip on my instance and ended up doing this. 😄 

Summary:

*   Added a lookup function to all Persistent Volume Claims (PVCs). This facilitates immutable fields during Helm upgrades and improves compatibility with plugins such as helm-diff, which is used by helmfile. Without the lookup, helmfile consistently detects changes and initiates upgrades. (Note: To fully utilize this, include `--dry-run=server` in your helmfile diff arguments.)
*   Implemented persistence support for symbolicator in deployment mode.
*   Enhanced geoip configuration options. The geoip volume necessitates ReadWriteMany access, which is often unavailable by default on managed Kubernetes services. For instance, in GKE, you can now deploy a fuse storage driver (object storage driver) and configure a volume for geoip using annotations and the extraManifests options.


mzglinski:
Hi @Mokto, as per https://github.com/sentry-kubernetes/charts/pull/1851#issuecomment-3244182098 do you want any changes here?

mzglinski:
I found another bug: the PVC used to have hooks, which caused it to be removed before every upgrade. This must have never worked, as you cannot delete a PVC while its consumers are running. Therefore, I don't believe we need to classify this as a breaking change.

mzglinski:
Hi @Mokto, we’re approaching the 25.10.0 release deadline, and I’m willing to submit a PR here, but I’d really appreciate it if this could be merged, since there will be some changes required to the Symbolicator PVC.

https://github.com/getsentry/self-hosted/issues/3985

Can you take a look at this?

charts/sentry/templates/relay/deployment-relay.yaml
```diff
@@ -212,7 +213,7 @@ spec:
       {{- if and .Values.geodata.volumeName .Values.geodata.accountID }}
       - name: {{ .Values.geodata.volumeName }}
         persistentVolumeClaim:
-          claimName: {{ .Values.geodata.volumeName }}
+          claimName: data-sentry-geoip
```
Mokto, line 216: That looks like a breaking change?

mzglinski: Not really, because it could never work with anything other than `data-sentry-geoip` since this is hardcoded in other places: https://github.com/search?q=repo%3Asentry-kubernetes%2Fcharts%20data-sentry-geoip&type=code

mzglinski: I decided to hardcode it here instead of updating the other places to use `.Values.geodata.volumeName`, because you cannot change the name of a PVC after its creation. I reasoned that switching to `.Values.geodata.volumeName` would introduce a breaking change.

Mokto:
Ok thanks!

PR #1862, opened 2025-09-03
fix(templates): fix indentation for multiple deployments yamls

Dentling:
The recently added labels are not correctly placed in some files.

This fixes this.


Mokto:
Thanks!

PR #1886, opened 2025-09-18
feat: Enhance clickhouse cleanup job

kimxogus:
1. [clickhouse-client](https://hub.docker.com/r/clickhouse/clickhouse-client/tags) image is not maintained(latest image is version 22, it doesn't support `DELETE FROM` syntax), so I changed it to clickhouse-server with configurable repository and tag
2. Default values.yaml doesn't set clickhouse password. In that config, you get error message `Bad arguments: the argument for option '--password' should follow immediately after the equal sign`, so I added `--password` arg optional
3. You also get error message `Code: 36. DB::Exception: Received from sentry-clickhouse:9000. DB::Exception: ALTER UPDATE/ALTER DELETE statements must use only deterministic functions. Function 'now' is non-deterministic. (BAD_ARGUMENTS)`, so I made the cut off timestamp in bash variable, and passed it to clickhouse query.
4. In discovering tables, I made it to filter tables having `timestamp` column only.

### DIFF
--- a/charts/sentry/templates/snuba/cleanup/cronjob-clickhouse-cleanup.yaml
+++ b/charts/sentry/templates/snuba/cleanup/cronjob-clickhouse-cleanup.yaml
@@ -69,8 +69,8 @@ spec:
           {{- end }}
           containers:
           - name: {{ .Chart.Name }}-clickhouse-cleanup
-            image: clickhouse/clickhouse-client:latest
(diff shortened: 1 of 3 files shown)


kimxogus:
@Mokto Could you review this? clickhouse cleanup job is not working in my test environment.

Mokto (approved):

PR #1888, opened 2025-09-19
fix: Add missing topics for snuba-subscription-consumer-eap-items

kimxogus:
snuba-subscription-consumer-eap-items failed because of these missing topics


Mokto:
Thanks.

Bafff:
Hey! Can somebody help me to understand would it help with receiving spans v2 without transactions? Or this is not for that?

PR #1962, opened 2025-12-03
feat: allow to override livenessProbe & readinessProbe

Poil:
I need to override readinessProbe & livenessProbe on my platform, actually sometimes we can sometimes not

PS: I think we should use port name, not an include/template for probe & co


Mokto:
:wave: Hi, @Poil,

PR #1970, opened 2025-12-09
fix: snuba-subscription-consumer-eap-items followed-consumer-group

MarleDK:
I noticed when looking at metrics in our deployment that `snuba_subscriptions_scheduler_executor_followed_group_mismatch{dataset="events_analytics_platform",group="eap_items_group"}` was happening a lot.
It seems to be because of this misconfig.

It looks like a minor oversight from the [initial upgrade to Sentry 25.8.0](https://github.com/sentry-kubernetes/charts/pull/1851).

This also matched the self-hosted: https://github.com/getsentry/self-hosted/blob/83f07124bf9f5a821cf854f625378da20a8cebdb/docker-compose.yml#L465


Mokto (approved):
thanks!

PR #1972, opened 2025-12-09
feat(ingress): Add native Traefik IngressRoute support

salehi:
# PR Title
feat(ingress): Add native Traefik IngressRoute support

## Description

This PR adds native Traefik `IngressRoute` CRD support as an alternative to standard Kubernetes `Ingress` resources.

### Problem

When using Traefik as the ingress controller, the current `Ingress` resource with regex paths does not work correctly. The chart generates:

```yaml
spec:
  ingressClassName: traefik
  rules:
  - host: sentry.example.com
    http:
      paths:
      - path: /api/[1-9][0-9]*/(.*)
        pathType: ImplementationSpecific
        backend:
          service:
            name: sentry-relay
            port:
              number: 3000
```

With nginx annotations:
```yaml
annotations:
  nginx.ingress.kubernetes.io/use-regex: "true"
```

**The problem:** Traefik ignores nginx-style annotations, so the regex paths are not matched correctly. All traffic falls through to the `/` catch-all route and is sent to `sentry-web` instead of `sentry-relay`.

This causes SDK event submission to fail with:
- 403 CSRF errors (hitting web UI login instead of relay)
- 302 redirects to login page

### Solution

Add native Traefik `IngressRoute` CRD support that properly handles regex path matching:

```yaml
apiVersion: traefik.io/v1alpha1
kind: IngressRoute
spec:
  routes:
    - match: Host(`sentry.example.com`) && PathRegexp(`/api/[0-9]+/`)
      kind: Rule
      priority: 100
      services:
        - name: sentry-relay
          port: 3000
```

### Changes

1. **`values.yaml`**: Added `ingressRoute` configuration section
2. **`templates/ingressroute.yaml`**: New template for Traefik IngressRoute CRDs
3. **`templates/_helpers.tpl`**: Added `relay.port` helper (if not present)
4. **`README.md`**: Documentation for Traefik IngressRoute usage

### Usage

```yaml
ingress:
  enabled: false  # Disable standard Ingress

ingressRoute:
  enabled: true
  hostname: sentry.example.com
  entryPoints:
    - web
    - websecure
  # tls:
  #   secretName: sentry-tls  # Optional
```

### Testing

Tested with:
- Traefik v2.x / v3.x on k3s
- Sentry 25.9.0
- Self-hosted Sentry Helm chart 27.8.0

Verified:
- `/api/2/store/` returns 401 (relay responding) instead of 302 (web redirecting)
- SDK events successfully ingested
- Web UI accessible at `/`
- Relay-to-relay communication working at `/api/0/relays/`

### Related Issues

- Fixes routing issues when using Traefik as ingress controller

### Checklist

- [x] Code follows project style guidelines
- [x] Documentation updated
- [x] Tested with Traefik ingress controller
- [x] Backwards compatible (new feature, opt-in)

### DIFF
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -2401,7 +2401,7 @@ ingress:
   # - secretName:
   #   hosts:
 
-ingressRoute:
+traefikIngressRoute:
   # -- Enable Traefik IngressRoute instead of standard Ingress
   # Use this when running Traefik as your ingress controller for proper regex path matching
   enabled: false
--- a/charts/sentry/values.yaml
+++ b/charts/sentry/values.yaml
@@ -2401,6 +2401,36 @@ ingress:
   # - secretName:
   #   hosts:
 
+ingressRoute:
+  # -- Enable Traefik IngressRoute instead of standard Ingress
+  # Use this when running Traefik as your ingress controller for proper regex path matching
+  enabled: false
+
+  # -- Traefik entrypoints to use
+  entryPoints:
+    - web
+    - websecure
+
+  # -- Hostname for the IngressRoute
+  # hostname: sentry.example.com
+
+  # -- TLS configuration (optional, omit if TLS is terminated upstream)
+  tls: {}
+    # secretName: sentry-tls
+    # options:
+    #   name: default
+    #   namespace: default
+
+  # -- Additional route annotations
+  annotations: {}
+
+  # -- Additional route labels
+  labels: {}
+
+  # -- Route priority (higher = matched first)
+  priority: 100
+
+
 filestore:
   # Set to one of filesystem, gcs or s3 as supported by Sentry.
   backend: filesystem
--- a/charts/sentry/templates/_helpers.tpl
+++ b/charts/sentry/templates/_helpers.tpl
@@ -1,0 +1,8 @@
+# Add this to templates/_helpers.tpl if not already present
+
+{{/*
+Relay port
+*/}}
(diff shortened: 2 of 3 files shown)


patsevanton:
Why traefik? Maybe add gateway API?

cynecx:
We've had issues because our cluster uses traefik v3 and the rule syntax has changed. The workaround for that was to fallback to the v2 rulesyntax:

```
ingress:
  enabled: true
  regexPathStyle: traefik
  ingressClassName: traefik
  annotations:
    # Make the ingress work with Traefik v3
    traefik.ingress.kubernetes.io/router.rulesyntax: v2
```

This seems to work quite well for us. The obvious caveat is that we don't know how long they are gonna keep the fallback working and not fully deprecate it.

salehi:
> Why traefik? Maybe add gateway API?

I just deployed to k3s, the default it traefik.

salehi:
I'm using these changes for compatibility with k3s and just wanted to share the work with the main stream.

charts/sentry/values.yaml
```diff
@@ -2401,6 +2401,36 @@ ingress:
   # - secretName:
   #   hosts:
 
+ingressRoute:
```
Mokto: I think you should call it "traefikIngressRoute" ?

salehi: Yes, you are right.
Done.

salehi:
force pushed for rebase from develop.

Mokto (approved):

PR #1981, opened 2025-12-19
feat!: upgrade to v25.12.1

mzglinski:
Significant changes have been made, and I intend to further refine the documentation & ci configuration. However, due to the holiday season, I will be unable to dedicate time to this for several weeks. Therefore, I'm releasing it now for those eager to explore!

This is a reflection of changes from here: https://github.com/getsentry/self-hosted/compare/25.8.0...25.12.0

Additionally, I've implemented taskBroker and taskWorker using the architecture recommended by [aldy505](https://github.com/aldy505)

The decision to not update the clickhouse chart is the result of this [conversation](https://github.com/sentry-kubernetes/charts/issues/1828)

All feedback is welcome. While I've tested these changes and they appear to function correctly on my end, Sentry offers a vast array of features. I encourage anyone willing to test their configuration against these updates to do so.


README.md
```diff
@@ -18,6 +18,25 @@ helm install my-sentry sentry/sentry --wait --timeout=1000s
 
 For now the full list of values is not documented, but you can get inspired by the `values.yaml` specific to each directory.
 
+## Upgrading to Chart 28.x.x
+
+### Storage Configuration Changes
+
+This release introduces significant changes to how Sentry handles storage for `nodestore` (raw events) and `profiling`. We strongly recommend using an external S3-compatible storage provider (e.g., AWS S3, Google Cloud Storage, MinIO) for these components to ensure performance and scalability.
```
aldy505: One thing to note for AWS S3 and GCS users, `nodestore` is very write heavy (if you have tons of throughput), make sure this won't set your cloud bills to the roof.

README.md
```diff
@@ -18,6 +18,25 @@ helm install my-sentry sentry/sentry --wait --timeout=1000s
...
+
+- **Nodestore**: You can now configure S3-based node storage via `nodestore.s3`.
+- **Profiles**: The `filestore.profiles` section now supports an S3 backend. Using the `filesystem` backend is discouraged for production environments.
```
aldy505, line 28: Technically you can separate storage for file (avatars, debug files) and session replays too if you want.

charts/sentry/templates/sentry/_helper-sentry.tpl
```diff
@@ -630,3 +706,56 @@ sentry.conf.py: |-
...
+{{- define "sentry.initContainer.nodestore-s3" -}}
+{{- if .Values.nodestore.backend }}
+- name: install-nodestore-s3
+  image: "{{ template "sentry.image" . }}"
+  imagePullPolicy: {{ default "IfNotPresent" .Values.images.sentry.pullPolicy }}
+  command:
+    - sh
+    - -c
+    - |
+      pip install --target=/sentry-plugins https://github.com/getsentry/sentry-nodestore-s3/archive/main.zip
```
aldy505, line 729: Nice! I was worried you'd use the one from `github.com/stayallive/sentry-nodestore-s3`

charts/sentry/values.yaml
```diff
@@ -310,124 +323,108 @@ sentry:
...
+        consumerGroup: taskworker-ingest
+        replicas: 1
+      - name: long
+        topic: taskworker-long
+        consumerGroup: taskworker-long
+        replicas: 1
+      - name: products
+        topic: taskworker-products
+        consumerGroup: taskworker-products
+        replicas: 1
```
aldy505, line 362: Nice. Although usually `taskworker` and `taskworker-ingest` are enough.

aldy505 (commented):
Overall, this is a nice draft.

aldy505:
Oh right you want to hold this off. 25.12.1 will come on Monday.

README.md
mzglinski: We do have support for that in chart already, this just lists the new things.

charts/sentry/templates/sentry/_helper-sentry.tpl
mzglinski: Yeah, I did catch that. It would be nice if there were git tags in this repo so we can version this package and update it with Renovate.

charts/sentry/values.yaml
mzglinski: That's probably true. The nice thing about this implementation is that it is completely configurable by the user via the values of the chart.

mzglinski:
No worries, I won't finish this by Monday ;)

charts/sentry/templates/sentry/taskworker/deployment-taskworker.yaml
```diff
@@ -0,0 +1,184 @@
...
+      - name: taskworker
+        image: "{{ template "sentry.image" $root }}"
+        imagePullPolicy: {{ default "IfNotPresent" $root.Values.images.sentry.pullPolicy }}
+        command: ["sentry"]
+        args:
+          - "run"
+          - "taskworker"
+          - "--concurrency={{ default $root.Values.sentry.taskWorker.concurrency $worker.concurrency }}"
+          - "--rpc-host={{ template "sentry.fullname" $root }}-taskbroker-{{ $worker.brokerName }}:50051"
+          - "--num-brokers={{ default $root.Values.sentry.taskBroker.replicas $worker.brokerReplicas }}"
```
manishrawat1992, line 99: @aldy505 @mzglinski do the taskbrokers implement consumer balancing also?

aldy505: Yes.

charts/sentry/values.yaml
```diff
@@ -108,10 +113,18 @@ vroom:
...
+    ## you MUST include "ReadWriteMany" to allow both vroom and ingest-profiles pods
+    ## to mount the same volume. Example:
+    ## accessModes:
+    ##   - ReadWriteMany
     accessModes:
       - ReadWriteOnce
     # storageClassName: standard
     size: "10Gi"
+    # this will be passed to the SENTRY_BUCKET_PROFILES env var
+    bucketString: "file:///var/vroom/sentry-profiles"
```
alviss7, line 127: I think you need to put `file:///var/lib/sentry/files/profiles` to match `filestore.profiles.filesystem.path`.

Mokto:
:wave: Hi, @mzglinski,

charts/sentry/values.yaml
mzglinski: Thanks, I've actually changed it the other way around to maintain compatibility with the previous setting.

charts/sentry/docs/external-clickhouse.md
Mokto, line 1: Great!

Mokto (approved):

mosesdd:
Why is clickhouse chart still enabled by default as a dependency when it's not working with this sentry version anyway?

mzglinski:
Didn't give that much thought. You have to make significant changes either way. At least now the existing resources are not automatically deleted when someone tries the upgrade without reading the release notes. In the future, those charts will be disabled by default and then removed.

arianf:
> The decision to not update the clickhouse chart is the result of this https://github.com/sentry-kubernetes/charts/issues/1828
> 
> All feedback is welcome

I’m a bit confused by the rationale for deprecating the in-chart ClickHouse based on the Bitnami announcement. #1828 seems focused on Bitnami-provided chart dependencies moving/being deprecated (redis/postgresql/kafka/zookeeper/etc.).

ClickHouse dependency here is from `sentry-kubernetes/charts` (not Bitnami), and the chart values uses the upstream `clickhouse/clickhouse-server` image, which doesn't seem affected by the Bitnami announcement. Am I missing something here?

https://github.com/sentry-kubernetes/charts/blob/b70306ca35b96b5e6a13358ae8988255f49be045/charts/clickhouse/values.yaml#L122

Mokto:
The discussion on bitnami is what motivated us to reevaluate the burden of maintaining sub charts that were anyways not a good fit for any kind of production related Sentry.

There are only a few people working on the chart and our time is really limited + this is not official from sentry. I would rather have us focus on maintaining/upgrading sentry components.

The migration has been well explained by the PR author so I think we’re good.

The next databases should probably follow the same logic soon: Postgres/redis/kafka.

And we probably need to find another solution for nginx (bitnamilegacy).

PR #2009, opened 2026-01-13
fix: optionally take app_id and app_name from secrets for Github

paragor:
We have many Sentry installations, and a separate GitHub application is used for each. It will be useful to take appId and appName from the secret, as it's done for other fields.

### DIFF
--- a/charts/sentry/templates/sentry/_helper-sentry.tpl
+++ b/charts/sentry/templates/sentry/_helper-sentry.tpl
@@ -706,6 +706,12 @@ sentry.conf.py: |-
   ##########
   # Github #
   ##########
+  {{- if .Values.github.existingSecretAppIdKey }}
+  SENTRY_OPTIONS['github-app.id'] = os.environ.get("GITHUB_APP_ID")
+  {{- end }}
+  {{- if .Values.github.existingSecretAppNameKey }}
+  SENTRY_OPTIONS['github-app.name'] = os.environ.get("GITHUB_APP_NAME")
+  {{- end }}
   SENTRY_OPTIONS['github-app.private-key'] = os.environ.get("GITHUB_APP_PRIVATE_KEY")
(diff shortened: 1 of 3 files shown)


charts/sentry/templates/sentry/_helper-sentry.tpl
```diff
@@ -706,6 +706,13 @@ sentry.conf.py: |-
...
   # Github #
   ##########
+  {{- if not .Values.github.appId }}
+  SENTRY_OPTIONS['github-app.id'] = os.environ.get("GITHUB_APP_ID")
+  {{- end }}
+  {{- if not .Values.github.appName }}
+  SENTRY_OPTIONS['github-app.name'] = os.environ.get("GITHUB_APP_NAME")
+  {{- end }}
+  SENTRY_OPTIONS['github-app.private-key'] = os.environ.get("GITHUB_APP_PRIVATE_KEY")
   SENTRY_OPTIONS['github-app.private-key'] = os.environ.get("GITHUB_APP_PRIVATE_KEY")
```
Mokto: duplicate

charts/sentry/templates/sentry/_helper-sentry.tpl
```diff
@@ -706,6 +706,13 @@ sentry.conf.py: |-
   ##########
   # Github #
   ##########
+  {{- if not .Values.github.appId }}
```
Mokto: The conditions here seem wrong. Maybe check for existingSecretAppIdKey ?

charts/sentry/templates/sentry/_helper-sentry.tpl
paragor: Thanks! 
Fixed

charts/sentry/templates/sentry/_helper-sentry.tpl
paragor: You are right!
Fixed

Mokto (approved):

PR #2012, opened 2026-01-15
feat!: replace nginx chart, refactor ingress, unify routing configurations

mzglinski:
Bitnami nginx chart is deprecated, and the nginx solution was just a bottleneck anyway. Ingress was already enabled by default, and I cannot imagine anyone running this chart on pure nginx.

Changes here align nginx ingress configuration with that of the nginx container, with support for other types of ingress. Those, however, might require adding dedicated annotations by the user (like for the proxy configuration).


Mokto:
I like the idea of this (simplification) but I'm afraid we make it too complicated for other type of ingress gateways. Traefik, Kong, etc...

Traefik is already handled via IngressRoute but I'm afraid for other providers.

mzglinski:
I do not see how this change makes things more complicated for users. For example, previously proxy settings were configured at the NGINX level, but they also had to be set at the ingress level to maintain continuity. I would expect users to apply similar changes for other ingress providers as well.

This means that the only real change is the source used for the “reverse-engineering”: previously it was the NGINX configuration, and now it would be the ingress configuration.

In the worst-case scenario, users can still resolve this by disabling ingress (either entirely or selectively) and creating their own object that is fully compatible with whichever provider they are using. This is exactly what I was doing before I realized the chart supported a pure (at least partially) ingress-based solution.

Mokto:
No they didn't have to be set at the ingress level at all. We just needed to forward all queries to nginx.

That's the config I had before.

mzglinski:
And you did not encounter this error?

https://github.com/getsentry/self-hosted/issues/1927

The fix in this chart was only suggested via comments. https://github.com/sentry-kubernetes/charts/commit/4c994b6482adf1519cc8ee11b1611d1057cde893

Mokto:
I didn't know.

Don't get me wrong, I really like this PR and I think it's a move in a right direction. I just think it would be nice to support a few more providers. Maybe we can ask the help of people using other ingress gateways.

mzglinski:
That is a good idea. I, however, have never worked with anything other than Nginx ingress and do not have the infrastructure to test with other providers. How would you like to handle this CTA? Should I add it to the upgrade notes or via a pinned issue?

README.md
```diff
@@ -18,6 +18,14 @@ helm install my-sentry sentry/sentry --wait --timeout=1000s
 
 For now the full list of values is not documented, but you can get inspired by the `values.yaml` specific to each directory.
 
+## Upgrading to Chart 29.x.x
+
+The nginx container has been removed because it became a throughput bottleneck; routing directly via Kubernetes Ingress improves performance and reduces hop latency.
+
+- Remove any `nginx.*` values and ensure `ingress.enabled=true`.
```
Mokto: or `traefikIngressRoute.enabled=true`

README.md
```diff
@@ -18,6 +18,14 @@ helm install my-sentry sentry/sentry --wait --timeout=1000s
...
+- Configure `ingress.assets` and `ingress.static` to keep asset rewrites and response header handling.
```
Mokto: is that really needed ? If using nginx-ingress-controller?

README.md
```diff
@@ -18,6 +18,14 @@ helm install my-sentry sentry/sentry --wait --timeout=1000s
...
+- If you previously relied on nginx extraLocationSnippet, move the logic to ingress annotations, your controller ConfigMap, or create dedicated ingress object manually or via `extraManifests`.
+
```
Mokto, line 36: Maybe add some text here that for now traefik & nginx-ingress-controller are supported but we welcome PRs?

Mokto:
Thanks for the reminder. Could you rebase this one please?

Mokto:
:wave: Hi, @mzglinski,

mzglinski:
@Mokto done!

Mokto:
Actually I've just merged this: https://github.com/sentry-kubernetes/charts/commit/1c9a2f9c8501a0a140b49d62525705e193afcece

So I think we should update the docs accordingly

jiriks74:
I'll have to look into this further but we need to have NGINX. We have some LUA scripts that wee need to have deployed on it. AFAIK it's possible to turn it off anyway so I don't really see the benefit of getting rid of it completely.

Also CloudPirates that are now used for `memcached` (https://github.com/sentry-kubernetes/charts/pull/2014) do have a Helm chart for NGINX that even uses the upstream container which would be easier for users to modify if neccessary:
https://github.com/CloudPirates-io/helm-charts/tree/main/charts/nginx/templates

mzglinski:
@jiriks74 It's a bit confusing, especially since you then have two ways to configure ingress, via this chart or via `nginx.ingress`. I think you should be able to achieve the same with pure ingress using `nginx.ingress.kubernetes.io/configuration-snippet` annotation

mzglinski:
@Mokto I've moved everything related to traffic routing into the chart README, leaving only a note and a reference in the main repo README.

Mokto:
> I'll have to look into this further but we need to have NGINX. We have some LUA scripts that wee need to have deployed on it. AFAIK it's possible to turn it off anyway so I don't really see the benefit of getting rid of it completely.
> 
> Also CloudPirates that are now used for `memcached` (#2014) do have a Helm chart for NGINX that even uses the upstream container which would be easier for users to modify if neccessary: https://github.com/CloudPirates-io/helm-charts/tree/main/charts/nginx/templates


1. The chart/container is now deprecated 
2. We've talked about previously about removing dependencies on subcharts because we don't have a lot of man power to maintain every one of them.

mzglinski:
@jiriks74 Can you confirm that this will be sufficient for you? An example od lua script in nginx ingress: 

You have to set `controller.config.annotations-risk-level: Critical` if you are using the nginx ingress helm chart to deploy it.

```
annotations:
    nginx.ingress.kubernetes.io/configuration-snippet: |
      access_by_lua_block {
        local headers = ngx.req.get_headers()
        ngx.header["X-Custom-Header"] = "custom-value"
      }
```

I do not want this PR getting blocked because of this.

jiriks74:
> I do not want this PR getting blocked because of this.

Understandable, however it seems that replacing NGINX will prove to be quite difficult for us.

I'm looking into it as I'd like to make the setup more streamlined as well but the environment I'm working with has some custom things that limit me in using Ingress. I generally don't like custom things but in this case it allows us to have quite robust fail over including Sentry which is not possible with Ingress.

I'm looking into some options if this PR goes through but it seems that it would be basically migrating the Bitnami chart to eg. CloudPirates replicating what is about to be deleted.

From what I gathered from the current setup doing just an if statement that would disable ingress and leaving the NGINX configmap in is enough.
This would allow you tol strip out NGINX from Ingress simplifying it with a statement that people that want to use NGINX will have to do their own Ingress/servives.

I don't feel like that solution would be a significant maintenance burden since the if statement just disables everything, the NGINX chart is made by someone else, the chart uses the upstream NGINX image and the configmap is basically a copy of the sentry-selfhosted one as far as I can see.

> You have to set controller.config.annotations-risk-level: Critical if you are using the NGINX ingress helm chart to deploy it.

Sadly that seems to not be enough, though I managed to find an alternative to running LUA in NGINX. However, like mentioned above, I've found bigger issues since looking into this.

mzglinski:
@jiriks74 I'm still not convinced that what you need cannot be achieved on pure ingress since this is also an nginx container after all. But if this is indeed the case, and you do have some custom configuration, you can always deploy the nginx chart alongside the Sentry in the same namespace. Does this sound like an acceptable solution to you?

Having nginx and ingress led to confusion and misconfiguration between them. I really do not think keeping the nginx in-house has benefits for a large enough audience to justify that.

speedo17:
@mzglinski  Hi, 
this could be a deeper issue - do you know, `nginx-ingress` itself is also being deprecated in a month?
https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/

In reality that means getting rid of nginx altogether, which will block @jiriks74 in using nginx lua scripts.

mzglinski:
@speedo17 Yes, I do know, and a replacement for this has already been implemented.

https://github.com/sentry-kubernetes/charts/commit/1c9a2f9c8501a0a140b49d62525705e193afcece

We will most likely switch over to Gateway API as the default in a future release.

However:

- nginx isn't the only ingress controller out there
- Using a single Nginx pod as the router for Sentry is not a good option either, but it's something most users ended up with because it was the default.

If anyone requires some custom logic in front of Sentry, they can always deploy their own Nginx, Apache, or another service to do so. That does not mean that we should include one by default, creating a bottleneck.

mzglinski:
@jiriks74 @speedo17 @Mokto

I've thought a bit more about the changes proposed here, researched more into different ingress controllers, and the reasons for the presence of assets and static rewrites in the Sentry repo.

Here is my new proposal:

- cleaner and more transparent implementation of ingress object, dropped rewrites for assets and static as they no longer seem needed. Also dropped support for beta ingress, Kubernetes v1.19 was released on 26.08.2020, I think everyone should be updated at this point ;)
- extended documentation regarding different routing options, with a recommended setup for custom proxy service (if required)
- aligned configurations (to the best of possibilities) for traefikIngressRoute, Ingress and HTTPRoute (there were some differences between them)

Nginx ingress is still the enabled by default configuration, but it can be easily swapped by the user and us in the future.

Please review this version!

jiriks74:
> I'm still not convinced that what you need cannot be achieved on pure ingress since this is also nginx container after all.

Like I mentioned I may have got an alternative solution for the Lua scripts. The main issue that I found since looking into this is that setting up Ingress in our environment is quite hard as it's EOL in k8s and an internal solution is preferred as it allows us to have eg. cross-cluster services which Ingress cannot do as far as I know.

> you can always deploy the nginx chart alongside the Sentry in the same namespace... Does this sound like an acceptable solution to you?

That is possible though I'd prefer not to keep a core part of Sentry in sync across 2 places manually (I consider it core since it's still the entry point with `self-hosted`) including having be careful about naming things as things like `{{ sentry.fullName }}` couldn't be used in the chart.

> Having nginx and ingress led to confusion and misconfiguration between them. I really do not think keeping the nginx in-house has benefits for a large enough audience to justify that.

I get that it's confusing. I'm still looking into this trying to figure out the most simple and sane option. That's why I've offered the solution that would make Ingress and NGINX mutually exclusive with a warning next to NGINX that it should be only used when necessary and with a good reason and that when enabled, you'll have to manage Ingress/services yourself.

It could also be places at EOF next to PGBouncer so that users find Ingress before NGINX.

Do you think that these 3 things would make it the support conceivable?

---

PS: I'm really thankful that you're willing to have this conversation with me. I understand that I'm holding this back a bit so I really appreciate that you're trying to figure out what's best with me.

mzglinski:
Ok @jiriks74 I've reintroduced the nginx chart, but made it disabled by default, as are all of the routing options. I hope this will force users to read the documentation and make an informed decision about which routing option to use.

README.md
```diff
@@ -336,41 +354,6 @@ nginx:
     tls: true
 ```
 
-### Gateway API
```
mzglinski, line 339: This is a changelog entry and should not have been updated in the first place. This is why I reverted this change.

Mokto:
Thank you! More documentation is great and we achieve the initial goal that is removing the deprecated subchart

Issue #2002, opened 2026-01-12
28.0.1 Upgrade snuba migrations migrate Error use Chart Clickhouse Version 23.8.16.16


blunt1973:
### Issue submitter TODO list

- [ ] I've searched for an already existing issues [here](https://github.com/sentry-kubernetes/charts/issues)

### Describe the bug (actual behavior)

sentry-snuba-migrate job has an error:
```
2026-01-11 12:22:55,540 Reading column value
Traceback (most recent call last):
  File "/.venv/bin/snuba", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/snuba/snuba/cli/migrations.py", line 115, in migrate
    runner.run_all(
  File "/usr/src/snuba/snuba/migrations/runner.py", line 215, in run_all
    pending_migrations = self._get_pending_migrations()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/snuba/snuba/migrations/runner.py", line 516, in _get_pending_migrations
    group_migrations = self._get_pending_migrations_for_group(group)
[...]

PR #2013, opened 2026-01-20
Fix issue #2002 broken dependency on clickhouse

titanlien:
## Root Cause

The error occurred because Snuba migrations were trying to create a ClickHouse table column with the syntax `JSON(max_dynamic_paths=128)`, which is only supported in ClickHouse version 24.8 and later. The chart was using ClickHouse version 23.8.16.16, which doesn't support this syntax.

## Fix Applied

Updated the ClickHouse version in `charts/clickhouse/Chart.yaml`:

- __Before__: `appVersion: "23.8.16.16"`
- __After__: `appVersion: "24.8.7.41"`

This upgrade enables support for the `JSON(max_dynamic_paths=128)` syntax required by the Snuba EAP (Events Analytics Platform) migrations.

### DIFF
--- a/charts/clickhouse/Chart.yaml
+++ b/charts/clickhouse/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "23.8.16.16"
+appVersion: "24.8.7.41"
 description: ClickHouse is an open source column-oriented database management
   system capable of real time generation of analytical data reports using SQL
   queries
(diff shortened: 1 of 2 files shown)


Mokto:
I'm a bit confused by this PR as we removed support for Clickhouse in 28.

mzglinski:
And the "fix" is not as simple as just updating the appVersion. The chart is incompatible and would require more changes.

PR #2025, opened 2026-01-30
feat: add serviceAccount.shared option for using single ServiceAccount

paragor:
## Problem                                                                                                                                                                                                                         
                                                                                                                                                                                                                                     
When using AWS EKS Pod Identity to grant S3 access to Sentry components, the current chart creates a separate ServiceAccount for each component (e.g., `sentry-web`, `sentry-snuba`, `sentry-relay`). 
This requires:                                                                                                                                                                                                    
- Binding Pod Identity/IAM roles to multiple ServiceAccounts                                                                                                                                                                       
- Maintaining separate IAM role mappings for each ServiceAccount                                                                                                                                                                   
- Complex RBAC configuration                                                                                                                                                                                                       
                                                                                                                                                                                                                                     
For S3 access (used by filestore, profiles, and nodestore), it's more practical to use a single ServiceAccount across all Sentry components.                                                                                       
                                                                                                                                                                                                                                     
## Solution                                                                                                                                                                                                                        
                                                                                                                                                                                                                                     
Added a new `serviceAccount.shared` parameter that allows using a single ServiceAccount for all components:
- When `serviceAccount.shared: false` (default): maintains current behavior - each component gets its own ServiceAccount with a suffix (e.g., `sentry-web`, `sentry-snuba`)
- When `serviceAccount.shared: true`: all components use the ServiceAccount specified in `serviceAccount.name`

### DIFF
--- a/charts/sentry/templates/geoip/deployment-geoip-job.yaml
+++ b/charts/sentry/templates/geoip/deployment-geoip-job.yaml
@@ -15,7 +15,7 @@ spec:
       {{- end }}
     spec:
       {{- if .Values.serviceAccount.enabled }}
-      serviceAccountName: {{ .Values.serviceAccount.name }}-geoip
+      serviceAccountName: {{ if .Values.serviceAccount.shared }}{{ .Values.serviceAccount.name }}{{ else }}{{ .Values.serviceAccount.name }}-geoip{{ end }}
       {{- end }}
       initContainers:
         - name: init-create-geoip-dir
--- a/charts/sentry/templates/geoip/serviceaccount-geoip.yaml
+++ b/charts/sentry/templates/geoip/serviceaccount-geoip.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.geodata.accountID .Values.serviceAccount.enabled }}
+{{- if and .Values.geodata.accountID .Values.serviceAccount.enabled (not .Values.serviceAccount.shared) }}
 apiVersion: v1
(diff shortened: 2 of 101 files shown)


elafkaihi:
Hi @paragor, thanks for putting together this feature!

I've been working on a related PR ([#2062](https://github.com/sentry-kubernetes/charts/pull/2062)) addressing an issue I've run into when using IRSA/Pod Identity for S3 access. Specifically, in the current chart template, the S3 config in sentry.conf.py sets credentials defaults via:

```
'access_key': os.getenv("S3_ACCESS_KEY_ID", "")
'secret_key': os.getenv("S3_SECRET_ACCESS_KEY", "")
```

When using Pod Identity (with no static credentials), boto3 sees "" as explicit credentials and doesn't fall through to the IAM role token from Pod Identity. This breaks the credential provider chain for IRSA/Pod Identity users in AWS ( i think for Azure too ).

My PR ([#2062](https://github.com/sentry-kubernetes/charts/pull/2062)) wraps these blocks with conditionals, so if neither accessKey nor existingSecret is set, the lines are omitted from the config and boto3 can do its thing with the IAM role.

You mention in your description that "For S3 access … it's more practical to use a single ServiceAccount." Did you run into the same issue with the os.getenv default? 

I'd love to make sure I'm not missing something! Any feedback is appreciated :smiley: 

Thanks again!

Mokto:
:wave: Hi, @paragor,

Issue #2024, opened 2026-01-29
Clickhouse not working in version 25.3.6


lucasdk3:
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/sentry-kubernetes/charts/issues)

### Describe the bug (actual behavior)

hello, I have this error to up my sentry on clickhouse

/entrypoint.sh: neither CLICKHOUSE_USER nor CLICKHOUSE_PASSWORD is set, disabling network access for user 'default'
/entrypoint.sh: line 156: /etc/clickhouse-server/users.d/default-user.xml: Read-only file system

### values.yaml

sentry:
  clickhouse:
      enabled: true
      privileged: true
      nodeSelector: {}
      tolerations: []
      securityContext:
        enabled: true
        runAsUser: 101
        fsGroup: 101
      clickhouse:
        replicas: "1"
        imageVersion: 25.3.6
        privileged: true
        configmap:
          remote_servers:
            internal_replication: true
            replica:
              backup:
                enabled: false
          zookeeper_servers:
            enabled: true
            config:
              - index: "clickhouse"
                hostTemplate: "{{ .Release.Name }}-zookeeper-clickhouse"
                port: "2181"
          users:
            enabled: true
            # Em vez de deixar o entrypoint criar o arquivo, nós definimos aqui
            # para que o arquivo já nasça pronto no volume montado
            user:
              - name: default
                config:
                  password: ""
                  networks:
                    - ip: "0.0.0.0/0"
                  profile: "default"
                  quota: "default"
  
        persistentVolumeClaim:
          enabled: true
          dataPersistentVolume:
            enabled: true
            accessModes:
              - "ReadWriteOnce"
            storage: "30Gi"

### Helm chart version

28.1.0

### Steps to reproduce

edit values.yaml

### Screenshots

<img width="853" height="80" alt="Image" src="https://github.com/user-attachments/assets/0c52f95c-7f48-4016-8049-f0b01967c7bf" />

### Logs

/entrypoint.sh: neither CLICKHOUSE_USER nor CLICKHOUSE_PASSWORD is set, disabling network access for user 'default'
/entrypoint.sh: line 156: /etc/clickhouse-server/users.d/default-user.xml: Read-only file system

Schoomee:
The version of ClickHouse declared with the Helm Chart is not compatible with Sentry, even with the latest tag.

I went through a ClickHouse Operator.

Helm Chart for the operator:

[...]

PR #2029, opened 2026-02-03
chore!: remove bundled clickhouse chart

mzglinski:
For cleanup reasons + not removing it completely led to some confusion.

Examples:

- https://github.com/sentry-kubernetes/charts/pull/2013
- https://github.com/sentry-kubernetes/charts/issues/2024
- https://github.com/sentry-kubernetes/charts/issues/2002

---

Again, to avoid multiple major releases, should be bundled with https://github.com/sentry-kubernetes/charts/pull/2012

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ For now the full list of values is not documented, but you can get inspired by t
 
 ## Upgrading to Chart 29.x.x
 
+## External ClickHouse
+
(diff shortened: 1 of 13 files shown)


Mokto (approved):
Really nice to be able to simplify the project. 

Thanks!

PR #2037, opened 2026-02-04
feat(cronjob): add filesystem cleanup cronjob for filesystem backend

AKrugRu:
The cronjob is used to clean up the filesystem when the filesystem backend is enabled.
It uses the sentry.image image; alpine can be used instead, but there is no strong reason to introduce a new image.
The implementation is based on cronjob-sentry-cleanup.

### DIFF
--- a/charts/sentry/templates/cleanerfs/cronjob-cleanerfs.yaml
+++ b/charts/sentry/templates/cleanerfs/cronjob-cleanerfs.yaml
@@ -1,0 +1,97 @@
+{{- if .Values.cleanerfs.enabled }}
+apiVersion: {{ include "sentry.batch.apiVersion" . }}
+kind: CronJob
+metadata:
+  name: {{ template "sentry.fullname" . }}-sentry-cleanerfs
+  labels:
+    app: {{ template "sentry.fullname" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  schedule: "{{ .Values.cleanerfs.schedule }}"
+  successfulJobsHistoryLimit: {{ .Values.cleanerfs.successfulJobsHistoryLimit }}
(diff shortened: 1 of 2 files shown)


Mokto:
Thanks

mzglinski:
That storage should never be deleted based on retention days. From Sentry Discord, by @aldy505

> You'd need 4 buckets:
> - nodestore
> - profiling
> - session-replays
> - filestore
>
> If your retention days is 90 days, then set everything other than filestore to have 90 days of retention time

but the same applies for the filesystem filestore backend.

In my opinion, this PR should be reverted or this fact clearly stated in a comment for this cronjob.

PS. @aldy505, regarding your recent question about missing documentation, this might be a candidate since this page does not explain this aspect https://develop.sentry.dev/self-hosted/production-enhancements/external-storage/ very clearly.

AKrugRu:
This cron job is disabled by default.

Currently, when using S3 or GCS, data cleanup is delegated to the policies of these storage systems. When using FileStorage, cleanup is not performed at all.
https://develop.sentry.dev/self-hosted/production-enhancements/external-storage/#sentry

This cleanup method is described here:
https://develop.sentry.dev/self-hosted/troubleshooting/sentry/#sentry-data-volume-not-being-cleaned-up

If you know of a setting that needs to be enabled, please share.

mzglinski:
The issue is that filestore contains things like user-uploaded media, source maps, etc., things that should not be deleted after the retention period. That is not the case for profiling and nodestore. I think the documentation you linked is not up to date.

mzglinski:
Also, if you are using filesystem storage, `sentry cleanup`, cleans it for you

> Important: sentry cleanup command won't delete files that is stored on an external storage such as GCS or S3. You will have to configure your own cleanup mechanism by utilizing your storage provider's object retention configuration. This should be set accordingly to the SENTRY_EVENTS_RETENTION_DAYS, although you can set it as a different value than what's set on the Docker Compose file.

https://develop.sentry.dev/self-hosted/production-enhancements/external-storage/

and sentry cleanup is supported in this chart.

AKrugRu:
Thanks for the docs. I’ll check how it works in practice

PR #2046, opened 2026-02-12
fix: add missing dnsPolicy to deployments

SquaredPotato:
Hello, we have an internal-only instance of Sentry running on a cluster which also serves our main DNS. This means that pods that don't have a DNS policy will try to resolve a URL (sentry.internaldomain) using the cluster coreDNS, which doesn't have that URL configured.

This seems to cause some deployments to fail, which in turn causes a bunch of internal Sentry errors. 

This is the first time I've edited a helm chart. I tried testing it on my own repo but I have absolutely no clue on how to make an actual helm repository and add it to the cluster, so I haven't been able to test it yet, but this seems to be what is going wrong in our case.

### DIFF
--- a/charts/sentry/Chart.yaml
+++ b/charts/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 29.2.0
+version: 29.2.1
 # renovate image=ghcr.io/getsentry/sentry
 appVersion: 26.1.0
 dependencies:
--- a/charts/sentry/Chart.yaml
+++ b/charts/sentry/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sentry
 description: A Helm chart for Kubernetes
 type: application
-version: 29.2.1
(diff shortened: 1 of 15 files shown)


SquaredPotato:
Alright hold on, I'll try to fix these linter issues in my fork

SquaredPotato:
Alright I'm back, figured out I could run the `ct` commands locally on my cluster to test it and it seems to work now :D

I'll first re-run the check with customized DNS settings, then add a commit to revert it back to default settings.

Mokto:
:wave: Hi, @SquaredPotato,

SquaredPotato:
I messed it up a lil bit there but if you squash my commits no-one will be any wiser

Mokto:
Could you document this in `values.yaml` please ?

SquaredPotato:
> Could you document this in `values.yaml` please ?

Hmmm sure, I mean it's existing functionality it just wasn't applied consistently across all deployments. I'll add a note

SquaredPotato:
Something like this?

Mokto (approved):

Mokto:
Thanks!

PR #2092, opened 2026-03-19
feat: replace bitnami pgbouncer image with icoretech/pgbouncer-docker…

shcherbak:
please review solution for replace outdated pgbouncer by bitnami with actual https://github.com/icoretech/pgbouncer-docker, all values and environments are preserved

Resolves https://github.com/sentry-kubernetes/charts/issues/1969

PR #2098, opened 2026-03-23
feat: replace bitnami pgbouncer image with icoretech/pgbouncer-docker

### LINKED ISSUE #1969: FeatureRequest: use https://github.com/icoretech/helm pgbouncer as dependency
because bitnami broke supply chain we need to look for alternatives

I am interested to implement migration to https://github.com/icoretech/helm pgbouncer chart

shcherbak:
please review solution for replace outdated pgbouncer by bitnami with actual https://github.com/icoretech/pgbouncer-docker, all values and environments are preserved

resolves https://github.com/sentry-kubernetes/charts/issues/1969

replaces pr https://github.com/sentry-kubernetes/charts/pull/2092

### DIFF
--- a/charts/sentry/templates/pgbouncer/pgbouncer-deployment.yaml
+++ b/charts/sentry/templates/pgbouncer/pgbouncer-deployment.yaml
@@ -22,45 +22,70 @@ spec:
         app: {{ template "sentry.fullname" . }}-pgbouncer
         {{- include "sentry.component.labels" (dict "component" "pgbouncer" "ctx" .) | nindent 8 }}
     spec:
+      initContainers:
+        - name: config
+          image: busybox
+          command: ["sh", "-c"]
+          args:
+            - |
+                cat <<EOF > /etc/pgbouncer/pgbouncer.ini
+                [databases]
+                ${PGBOUNCER_DATABASE}=host=${POSTGRESQL_HOST} port=${POSTGRESQL_PORT} dbname=${PGBOUNCER_DATABASE}
+                [pgbouncer]
+                listen_port=5432
+                listen_addr=0.0.0.0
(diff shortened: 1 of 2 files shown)


shcherbak:
@Mokto I've fixed conflicts, but one of commit messages not confirms with rules. Can You please remove non compliant one?

Mokto:
As explained in the discussion before, I'm not entirely sure about this.

The goal was to deprecate supporting those extra subcharts and enforce people to deploy a proper postgres/clickhouse/etc.

Also from a security point of view I'm not fund of using a github repository with 2 stars.

shcherbak:
> As explained in the discussion before, I'm not entirely sure about this.
> 
> The goal was to deprecate supporting those extra subcharts and enforce people to deploy a proper postgres/clickhouse/etc.
> 
> Also from a security point of view I'm not fund of using a github repository with 2 stars.

You are right, 2 stars is no good, but there is no official pgbouncer image, all industry lays on bitnami supply chain

Fabccc:
I agree with Mokto point of view on removing dependencies from the chart. 
Brief feedback on pgbouncer, we've been running the icoretech/pgbouncer helm chart in production for over a year now without any trouble. We have a relatively low usage deployment so it might not be representative of real-world deployment (1M Errors, 1M transactions, 100M spans monthly)

patsevanton:
@shcherbak @Fabccc May be use https://github.com/edoburu/docker-pgbouncer like here  https://github.com/getsentry/self-hosted/blob/master/docker-compose.yml#L178 ?

shcherbak:
@patsevanton @Mokto If You are agree to use edoburu/docker-pgbouncer I will adopt it to the manifest

PR #2102, opened 2026-03-27
feat!: separate livenessProbe & readinessProbe

mzglinski:
We should be able to configure them separately (there is no point of readinessProbe being the exact same as livenessProbe).


Mokto (commented):
Thanks. You're missing some upgrade notes

mzglinski:
Added.

mzglinski:
@Mokto this one is ready as well :)

PR #2121, opened 2026-04-15
fix: correctly capture sentry createuser exit code in user-create hook

InfraArtists:
The previous code used `|| true`, which reset `$?` to 0 before it could
be checked, making the exit code guard a no-op. Capture the exit code
immediately after the command and propagate the actual code on failure.
Without this fix, any non-zero exit code from sentry createuser (other than 3)
would silently be treated as success, masking failures.

### DIFF
--- a/charts/sentry/templates/hooks/user-create.yaml
+++ b/charts/sentry/templates/hooks/user-create.yaml
@@ -89,11 +89,12 @@ spec:
               --no-input \
               --superuser \
               --email "{{ .Values.user.email }}" \
-              --password "$ADMIN_PASSWORD" || true; \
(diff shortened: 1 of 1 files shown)


InfraArtists:
@Mokto, can you please review the fix?

Mokto:
so what's 3 for?

InfraArtists:
@Mokto 
3 = user already exists 
The root cause is `|| true`. It resets `$? `to `0` regardless of what the actual command returned, so the exit code check after it was effectively a no-op; it would always take the success path, even on real failures.

With `||true`, no matter which command you are running, the exit code is always 0.

Please have a look at this sample:
exit_code

Mokto:
Great that's what I wanted to make sure of !

PR #2126, opened 2026-04-21
feat(taskworker): add maxChildTaskCount value to prevent OOMKill from memory growth

ffrenzy71-arch:
## Problem

Taskworker child processes accumulate memory over time leading to OOMKill. This is a known issue reported in [getsentry/self-hosted#4265](https://github.com/getsentry/self-hosted/issues/4265).

The Sentry CLI already supports `--max-child-task-count` which restarts child processes after N tasks, preventing memory from growing unboundedly.

## Solution

Adds `maxChildTaskCount` field to `taskWorker` values (and per-worker override via `workers[].maxChildTaskCount`), passed as `--max-child-task-count` to the taskworker command.

The flag is optional — if not set, behaviour is unchanged.

## Usage

```yaml
sentry:
  taskWorker:
    maxChildTaskCount: 10000
```

Or per-worker:

```yaml
sentry:
  taskWorker:
    workers:
      - name: ingest
        brokerName: ingest
        maxChildTaskCount: 5000
```

## References

- Fix applied to self-hosted docker-compose: https://github.com/getsentry/self-hosted/pull/4279
- Issue: https://github.com/getsentry/self-hosted/issues/4265


Mokto (approved):

patsevanton:

PR #2139, opened 2026-04-29
move External ClickHouse to readme

patsevanton:
### Summary

Moves the External ClickHouse configuration guide from a separate doc file (`charts/sentry/docs/external-clickhouse.md`) into the main `README.md` for better visibility and discoverability.

### Motivation

The External ClickHouse setup documentation was located in a nested docs directory where users are less likely to find it. Since using an external ClickHouse is the recommended approach (due to the bundled Bitnami chart being considered legacy per [Issue #1828](https://github.com/sentry-kubernetes/charts/issues/1828)), this information should be prominently placed in the main README.


patsevanton:

PR #2151, opened 2026-05-03
fix!: require admin password and remove insecure default

InfraArtists:
## Summary

The chart currently ships to the user.password: aaaa` as the default admin password in [values.yaml](charts/sentry/values.yaml). This is a known-bad credential publicly exposed in version control.
Any deployment that doesn't explicitly override it is vulnerable from day one.

This PR:
- Removes the default password (sets `user.password: ""`)
- Adds a Helm template-time check in `templates/hooks/user-create.yaml` that fails the install with a clear, actionable error message if neither `user.password` nor `user.existingSecret` is set when `user.create` is true
- Updates the README quickstart to show both the `--set user.password=` and `--set user.existingSecret=` install paths

## Why a fail-fast check?

Previously, a user who forgot to override the password would silently inherit `aaaa`. Now they get a clear error in the terminal at `helm install`/`helm upgrade` time, before any resources are created:

Please have a look at the screenshot:

image

Note for reviewer: 
Happy to split this into two PRs (default removal + fail-fast validation) if preferred.
Also open to adjusting the error message wording or moving the check into a `_helpers.tpl` if that fits the project style better.


InfraArtists:
Hi @Mokto, sorry to bother you. Just a gentle ping to see if you've had a chance to review this. No rush, just wanted to make sure it didn't get lost in the notifications. Thanks

Mokto:
Could you please also change the upgrade notes?

InfraArtists:
@Mokto  Done.

Mokto (approved):
Thanks for the PR

mzglinski:
The upgrade notes amended the instructions for v30, and they should be in the newly created v31.

InfraArtists:
@mzglinski Sorry, my bad.

@Mokto, can you edit the `CHANGELOG.md` directly, or should I open new PR?

Mokto:
Could you open a PR please ?

InfraArtists:
@Mokto , yes sure, it's my pleasure. Here it is #2158

PR #2155, opened 2026-05-05
fix: harden shell scripts with set -euo pipefail

patsevanton:
## Summary

- Replace `set -e` with `set -euo pipefail` in Kafka provisioning and ClickHouse cleanup scripts to fail fast on unset variables and pipeline errors.

## Motivation

Using only `set -e` has known pitfalls:

- **Unset variables** silently expand to empty strings, causing silent bugs or dangerous commands (e.g., `rm -rf $UNSET_VAR/`).
- **Pipeline failures** — only the exit code of the last command is checked; earlier commands in a pipeline can fail silently.

`set -euo pipefail` addresses all three:

| Flag | Effect |
|------|--------|
| `-e` | Exit immediately if a command exits with non-zero status |
| `-u` | Treat unset variables as an error |
| `-o pipefail` | Pipeline returns the exit code of the last command that failed (not just the last command) |

### `templates/kafka-provisioning/configmap-kafka-provisioning.yaml`

- `wait-for-kafka.sh`: `set -e` → `set -euo pipefail`
- `provision.sh`: `set -e` → `set -euo pipefail`

### `templates/snuba/cleanup/cronjob-clickhouse-cleanup.yaml`

- ClickHouse cleanup script: `set -e` → `set -euo pipefail`

### DIFF
--- a/charts/sentry/templates/kafka-provisioning/configmap-kafka-provisioning.yaml
+++ b/charts/sentry/templates/kafka-provisioning/configmap-kafka-provisioning.yaml
@@ -42,7 +42,7 @@ metadata:
 data:
   wait-for-kafka.sh: |
     #!/bin/bash
-    set -e
+    set -euo pipefail
     {{- range $brokers }}
     timeout=120; until nc -z {{ .host }} {{ .port }} 2>/dev/null; do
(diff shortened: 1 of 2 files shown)


patsevanton:

Mokto (approved):

PR #2154, opened 2026-05-05
chore: release develop

Mokto:
:robot: I have created a release *beep* *boop*
---

<details><summary>sentry: 31.0.0</summary>

### ⚠ BREAKING CHANGES

* require admin password and remove insecure default ([#2151](https://github.com/sentry-kubernetes/charts/issues/2151))
* add support topicctl for external kafka ([#2157](https://github.com/sentry-kubernetes/charts/issues/2157)) ([9ae8784](https://github.com/sentry-kubernetes/charts/commit/9ae878467a6ab49367dd452e3ca15e8c2ea3aa08))

### Features

* add support topicctl for external kafka ([#2157](https://github.com/sentry-kubernetes/charts/issues/2157)) ([9ae8784](https://github.com/sentry-kubernetes/charts/commit/9ae878467a6ab49367dd452e3ca15e8c2ea3aa08))
* **cache:** add cache.enabled flag, deprecate sourcemaps.enabled ([#2156](https://github.com/sentry-kubernetes/charts/issues/2156)) ([0dc0ff7](https://github.com/sentry-kubernetes/charts/commit/0dc0ff76f612617986eb6410607faf9f4cbe6561))

### Bug Fixes

* add missing -- separator before consumer CLI options to prevent pod crashloops ([#2149](https://github.com/sentry-kubernetes/charts/issues/2149)) ([579ab05](https://github.com/sentry-kubernetes/charts/commit/579ab05d435246803ec9d49af9e1a47cf9aa1fed))
* harden shell scripts with set -euo pipefail ([#2155](https://github.com/sentry-kubernetes/charts/issues/2155)) ([d34dd65](https://github.com/sentry-kubernetes/charts/commit/d34dd65d1dfd2394b65422bf84bc286acdfef433))
* require admin password and remove insecure default ([#2151](https://github.com/sentry-kubernetes/charts/issues/2151)) ([c14b1bf](https://github.com/sentry-kubernetes/charts/commit/c14b1bf3b875fa4cfd59d179d55086ffbc17aa68))
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #2156, opened 2026-05-05
feat(cache): add cache.enabled flag, deprecate sourcemaps.enabled

patsevanton:
## Summary

- Add new `cache.enabled` flag to control Django CACHES (memcached) independently from sourcemaps
- Deprecate `sourcemaps.enabled` — it will be removed in a future release
- Add deprecation warning in NOTES.txt shown during `helm install/upgrade`

## Motivation

In Sentry source code, Django `CACHES` (memcached) is used by 60+ components: models, API authentication, ingest, integrations, quotas, alerts, and more. Previously, CACHES was tied to `sourcemaps.enabled`, which was misleading — the cache serves far more than just source maps.

## Changes

| File | Change |
|---|---|
| `values.yaml` | Added `cache.enabled: false` flag, added deprecation comment on `sourcemaps.enabled` |
| `_helper-sentry.tpl` | Updated condition: `or .Values.cache.enabled .Values.sourcemaps.enabled` |
| `Chart.yaml` | Updated memcached condition: `cache.enabled,sourcemaps.enabled` |
| `NOTES.txt` | Added deprecation warning when `sourcemaps.enabled` is used |
| `README.md` | Updated parameters table and Source Maps section |

## Backward Compatibility

- `sourcemaps.enabled: true` — continues to work as before, memcached is deployed
- `cache.enabled: true` — memcached is deployed, CACHES enabled for all Sentry components
- Both `false` — memcached is not deployed (current default behavior preserved)

## Usage

```yaml
# Recommended (new)
cache:
  enabled: true

# Deprecated (still works, shows warning)
sourcemaps:
  enabled: true
```

### DIFF
--- a/charts/sentry/Chart.lock
+++ b/charts/sentry/Chart.lock
@@ -14,5 +14,5 @@ dependencies:
 - name: nginx
   repository: oci://registry-1.docker.io/cloudpirates
   version: 0.7.0
-digest: sha256:b1db91770a93016ac2b0ad35c779b4a0f268cf65d06de9558ac3fbe03185035f
-generated: "2026-02-25T13:11:28.901973313+01:00"
+digest: sha256:db3349bfd8df7edf86fd67be5b6a4b369b3dd76a1c5548c860e855c89bcd8e4f
+generated: "2026-05-05T17:50:54.803225971+06:00"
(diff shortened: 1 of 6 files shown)


patsevanton:

Mokto (approved):

Mokto:
@patsevanton we already have a breaking change (major version) ongoing, maybe it's better to break behavior before I merge it ? 

https://github.com/sentry-kubernetes/charts/pull/2154

PR #2187, opened 2026-05-22
fix: add timeoutSeconds to exec livenessProbe in consumer deployments

roko99:
**Problem**

  48 consumer/worker deployment templates render exec-based livenessProbe blocks without a timeoutSeconds field:

```
  livenessProbe:
    exec:
      command:
        - rm
        - /tmp/health.txt
    initialDelaySeconds: 5
    periodSeconds: 320
    # timeoutSeconds missing → defaults to 1s
```
  When timeoutSeconds is omitted, Kubernetes defaults to 1s. This was harmless until now because Kubernetes had a long-standing bug where timeoutSeconds on exec probes was silently ignored since 1.0.

  Kubernetes 1.35 ([#134635](https://github.com/kubernetes/kubernetes/pull/134635)) locked the gate to true — enforcement is now mandatory on all distributions (GKE, EKS, AKS, etc.).

  Result: any exec probe without an explicit timeoutSeconds now runs with a hard 1s deadline.


patsevanton:

Mokto (approved):

PR #2191, opened 2026-05-22
feat: add concurrency, inputBlockSize, maxBatchSize, maxBatchTimeMs support for ingest-occurences consumer

spq1tr:
## Problem

`sentry-ingest-occurences` consumer was added to the chart without support for configuring consumer-specific CLI flags.

## Solution

Adds `concurrency`, `inputBlockSize`, `maxBatchSize`, and `maxBatchTimeMs` values for the consumer, passed as consumer-specific arguments after the `--` separator.

Flags were verified against the running binary:
```
$ sentry run consumer ingest-occurrences --consumer-group ingest-occurrences -- --help
Usage: ingest-occurrences [OPTIONS]

Options:
  --processes INTEGER
  --input-block-size INTEGER
  --output-block-size INTEGER
  --max-batch-size INTEGER        Maximum number of messages to batch before
                                  flushing.
  --max-batch-time-ms INTEGER     Maximum time (in milliseconds) to wait
                                  before flushing a batch.
  --mode [batched-parallel|parallel]
                                  The mode to process occurrences in. Batched-
                                  parallel uses batched in parallel to
                                  guarantee messages are processed in order
                                  per group, parallel uses multi-processing.
  --help                          Show this message and exit.
```

## Changes

- `templates/deployment-sentry-ingest-occurences.yaml` — emit `--` separator and conditionally pass `--processes`, `--input-block-size`, `--max-batch-size`, `--max-batch-time-ms`
- `values.yaml` — add commented-out `concurrency`, `inputBlockSize`, `maxBatchSize`, `maxBatchTimeMs` keys to the consumer section, matching the style of sibling consumers

### DIFF
--- a/charts/sentry/templates/sentry/ingest/occurrences/deployment-sentry-ingest-occurrences.yaml
+++ b/charts/sentry/templates/sentry/ingest/occurrences/deployment-sentry-ingest-occurrences.yaml
@@ -107,6 +107,25 @@ spec:
           - "--healthcheck-file-path"
           - "/tmp/health.txt"
           {{- end }}
+          {{- if or .Values.sentry.ingestOccurrences.concurrency .Values.sentry.ingestOccurrences.inputBlockSize .Values.sentry.ingestOccurrences.maxBatchSize .Values.sentry.ingestOccurrences.maxBatchTimeMs }}
+          - "--"
+          {{- end }}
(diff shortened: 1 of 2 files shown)


patsevanton:

Mokto (approved):

PR #2195, opened 2026-05-22
fix: correct ingestFeedback serviceAccount condition and remove duplicate snuba-replacer option

patsevanton:
### Changes

1. **Fix ingestFeedback ServiceAccount condition** (`serviceaccount-sentry-ingest-feedback.yaml`):
   - The ServiceAccount template was incorrectly gated by `sentry.ingestMonitors.enabled` instead of `sentry.ingestFeedback.enabled`, causing the feedback ingest ServiceAccount to be rendered based on the wrong feature flag.

2. **Remove `--no-strict-offset-reset` from snuba-replacer** (`deployment-snuba-replacer.yaml`):
   - The `--no-strict-offset-reset` flag is a duplicate of the default behavior and is no longer needed. Removing it simplifies the deployment configuration.


patsevanton:

patsevanton:

Mokto (approved):

PR #2197, opened 2026-05-23
fix: remove dead code prefix and fix incorrect comments in externalClickhouse

patsevanton:
#### 1. Remove unused `sentry.prefix` helper and `prefix` parameter (dead code)

The `sentry.prefix` helper in `_helper.tpl` and the corresponding `prefix:` parameter in `values.yaml` were never referenced by any template. Removed:
- Helper definition `sentry.prefix` from `templates/_helper.tpl`
- Parameter `prefix:` from `values.yaml` (line 1)
- Table row from `README.md`
- Usage examples from `docs/usage-aws-terraform.md` and `docs/usage-digitalocean.md`

#### 2. Fix incorrect comment in `externalClickhouse` section

The comment in `values.yaml:2866` incorrectly referenced `'postgres-password'` as the default secret key, but the actual default used in `templates/_helper.tpl:654` is `'clickhouse-password'`. Users following the old comment would create a secret with the wrong key name, causing silent authentication failures against ClickHouse.

#### 3. Fix outdated URL (`clickhouse.tech` → `clickhouse.com`)

The `clickhouse.tech` domain was retired and redirects to `clickhouse.com`. Updated the documentation link in the `externalClickhouse` section of `values.yaml`.

### DIFF
--- a/charts/sentry/README.md
+++ b/charts/sentry/README.md
@@ -400,7 +400,6 @@ Note: this table is incomplete, so have a look at the values.yaml in case you mi
 | postgresql.replication.numSynchronousReplicas | int | `1` |  |
 | postgresql.replication.readReplicas | int | `2` |  |
 | postgresql.replication.synchronousCommit | string | `"on"` |  |
-| prefix | string | `nil` |  |
 | redis.auth.enabled | bool | `false` |  |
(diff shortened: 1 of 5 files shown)


patsevanton:

Mokto (approved):

PR #2219, opened 2026-05-29
feat: enable cache (memcached) by default

patsevanton:
## What this PR does

Enables Django cache (memcached) by default, removing the need for  and  toggles.

## Changes

- Remove  condition from 
- Remove  from memcached dependency in 
- Remove  and  sections from 
- Remove deprecated warning from 
- Update  to reflect that cache is now always enabled

## Motivation

Memcached is required by 60+ Sentry components for caching. Having it disabled by default causes issues for most deployments.


patsevanton:
@Mokto may be enable the cache by default?

Mokto (approved):
fine with me ;)

PR #2131, opened 2026-04-26
fix(taskworker): extend liveness probe initial delay to prevent CrashLoopBackOff (#2128)

patsevanton:
## Summary

Increase `taskWorker.livenessProbe.initialDelaySeconds` from `10` to `30` to prevent `CrashLoopBackOff` on taskworker pods.

## Problem Description

Pods `sentry-taskworker-default` and `sentry-taskworker-products` were stuck in `CrashLoopBackOff`, restarting indefinitely after ~40 seconds of runtime.

| Pod | Status | Restarts |
|-----|--------|----------|
| `sentry-taskworker-default-*` | **CrashLoopBackOff** | 12+ |
| `sentry-taskworker-products-*` | **CrashLoopBackOff** | 11+ |

Other taskworkers (`ingest`, `long`) also experienced initial restarts but occasionally stabilized, while `default` and `products` never recovered.

## Root Cause

The taskworker container uses a file-based liveness probe:

```yaml
livenessProbe:
  exec:
    command:
      - rm
      - /tmp/health.txt
```

The container starts with `--health-check-file-path=/tmp/health.txt`, which *should* create this file during initialization, but in practice the file is created too late (or never in time).

With the previous settings:
- `initialDelaySeconds: 10`
- `periodSeconds: 10`
- `failureThreshold: 3`

…the container was killed after only **~40 seconds** (`10 + 3 × 10 = 40s`), which was not enough for the worker to fully initialize and create the health check file.

Kubelet events confirmed the failure:

```
Liveness probe failed: rm: cannot remove '/tmp/health.txt': No such file or directory
```

## Fix

Increase `initialDelaySeconds` from `10` to `30` in the `sentry.taskWorker.livenessProbe` block in `values.yaml`.

This gives taskworker containers sufficient time to:
1. Connect to the taskbroker gRPC endpoint
2. Spawn child worker processes
3. Create the `/tmp/health.txt` health check file

**Changed file:** `charts/sentry/values.yaml`

```diff
   taskWorker:
     ...
     livenessProbe:
-      initialDelaySeconds: 10
+      initialDelaySeconds: 30
       periodSeconds: 10
       timeoutSeconds: 5
```

## Related Issue

Fixes #2128

Issue #2128, opened 2026-04-25
CrashLoopBackOff for sentry-taskworker-default and sentry-taskworker-products


patsevanton:
## Environment

| Component | Version |
|---|---|
| Kubernetes | v1.33.3 (Yandex Cloud Managed) |
| Helm chart | sentry-30.1.0 |
| Sentry (app version) | 26.2.1 |

## Problem Description

Two taskworker pods are in `CrashLoopBackOff` and restarting indefinitely:

| Pod | Status | Restarts | Age |
|---|---|---|---|
| `sentry-taskworker-default-7f97885cc6-r2prw` | **CrashLoopBackOff** | 12 | 34m |
| `sentry-taskworker-products-6c89889458-6gwsc` | **CrashLoopBackOff** | 11 | 34m |

Meanwhile, other taskworkers are running stably (after initial restarts):

| Pod | Status | Restarts |
|---|---|---|
| `sentry-taskworker-ingest-58477f56b-hpccz` | Running | 3 |
| `sentry-taskworker-long-5df46f9b5d-nr8lj` | Running | 3 |

All taskbrokers are running normally:

| Pod | Status | Restarts |
|---|---|---|
| `sentry-taskbroker-default-0` | Running | 0 |
| `sentry-taskbroker-products-0` | Running | 0 |
| `sentry-taskbroker-ingest-0` | Running | 0 |
| `sentry-taskbroker-long-0` | Running | 0 |

## Root Cause

The liveness probe on all taskworkers is configured as follows:

```yaml
livenessProbe:
  exec:
    command:
    - rm
    - /tmp/health.txt
  initialDelaySeconds: 10
  periodSeconds: 10
  timeoutSeconds: 5
  failureThreshold: 3
```

Probe logic: the `rm /tmp/health.txt` command is considered successful if the file **exists** and was deleted (exit code 0). If the file does not exist — `rm` returns an error → probe failed.

The taskworker container starts with the `--health-check-file-path=/tmp/health.txt` flag, which **should** create this file, but **does not** (or does so too late).

After 3 failed probes (initialDelay 10s + 3 × periodSeconds 10s = **~40 seconds**), kubelet kills the container. The container shuts down gracefully via shutdown sequence and restarts, forming an infinite CrashLoopBackOff cycle.

## Events for pod `sentry-taskworker-default`

```
[...]

Issue #2003, opened 2026-01-12
Worker liveness probe fails with IndexError when using `sentry exec -c` with app.control.ping()


kyuwan95:
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/sentry-kubernetes/charts/issues)

### Describe the bug (actual behavior)

The default worker liveness probe configuration in the Sentry Helm chart fails with an `IndexError`, causing workers to restart continuously every 3 minutes.

## Root Cause

When executing the liveness probe command, `app.control.ping()` returns an empty list `[]`:
```bash
root@sentry-worker:/usr/src/sentry# sentry exec -c 'from sentry.celery import app; import os; dest="celery@{}".format(os.environ["HOSTNAME"]); result=app.control.ping(destination=[dest], timeout=5); print(result)'
[]
```

This causes `result[0]` to raise an `IndexError` because the list is empty.

## Investigation

The worker itself is running correctly and processing tasks successfully:
```
07:39:43 [INFO] celery.apps.worker: celery@sentry-worker-68bf4c78db-592z9 ready.
07:39:43 [INFO] celery.app.trace: Task sentry.tasks.schedule_auto_transition_issues_new_to_ongoing[...] succeeded in 0.034s
```

However, `app.control.ping()` executed via `sentry exec` runs in a **separate Python process** and cannot communicate with the actual worker process, resulting in an empty response

### Expected behavior

The liveness probe should successfully verify that the Celery worker is alive and responsive, returning a successful exit code (0) so that Kubernetes does not restart the pod.

### values.yaml

```
kafka:
  enabled: false
externalKafka:
  cluster:
  - host: "***"
    port: 9092
  - host: "***"
    port: 9092
  - host: "***"
    port: 9092
  
redis:
  enabled: false
externalRedis:
  host: "***"
  port: 6379
  password: ***

postgresql:
  enabled: false
externalPostgresql:
  host: "***"
  port: 5432
  username: ***
  password: ***
  database: sentry

clickhouse:
  enabled: true

zookeeper:
  enabled: true

rabbitmq:
  enabled: false

ingress:
  enabled: false

user:
  create: true
  email: "***"
  password: "***"

system:
  adminEmail: "***"

sentry:
  worker:
    replicas: 1
    livenessProbe:
      timeoutSeconds: 20
    logLevel: "DEBUG"
    logFormat: "human"
    resources:
    requests:
      cpu: 250m
      memory: 800Mi
    limits:
[...]

Issue #1874, opened 2025-09-08
sentry-worker should have readinessProbe


Poil:
Hi,

I think sentry-worker should have readinessProbe, we have a lot of container restart when scaling up.

If you are agree I can do a PR.

Regards

Mokto:
This issue is stale because it has been open for 30 days with no activity.

Mokto:
This issue was closed because it has been inactive for 14 days since being marked as stale.

PR #2221, opened 2026-05-29
refactor: move livenessProbe into a shared helper

vitka:
Consolidate the inline `livenessProbe:` block duplicated across consumer/worker deployment templates into a single helper (`sentry.livenessProbe.execHealthcheckFile`) in `_helper.tpl`. Allow defining probe fields other than hardcoded ones.

Set up a single place to evolve the probe definition for follow-ups (#1874, #2003, #2128, #2131).

## Motivation

The same probe block is inlined many times with only the value path differing. The current template also silently drops any probe field other than `initialDelaySeconds` / `periodSeconds` / `timeoutSeconds`, so `failureThreshold` etc. can't be set per-consumer despite being in the values schema.


Mokto:
:wave: Hi, @vitka,

PR #2227, opened 2026-05-30
feat!: replace deprecated distutils.strtobool and enable rustConsumer

patsevanton:
## Changes

- Remove deprecated `from distutils.util import strtobool` import (deprecated since Python 3.10, removed in 3.12)
- Replace `strtobool()` calls with `.lower() in ("true", "1", "yes")` for mail TLS/SSL configuration options
- Enable `rustConsumer` by default

## Motivation

`distutils` module is deprecated since Python 3.10 and removed in Python 3.12. Using it causes an `ImportError` on newer Python versions. This change replaces the `strtobool` usage with an equivalent inline check that works across all Python versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 The changelog below refers to the main `sentry` chart only.
 
+## Upgrading to Chart 32.0.0
+
+- **Rust Snuba consumer** is now enabled by default. Set `snuba.rustConsumer: false` to revert.
+- Removed deprecated `distutils.strtobool` from `sentry.conf.py` template — mail TLS/SSL now uses `.lower() in ("true", "1", "yes")`.
+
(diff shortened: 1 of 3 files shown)


patsevanton:
@shcherbak @elafkaihi @paragor @Poil @bhataprameya @mzglinski @spq1tr @ShellWen @zacharyarnaise @InfraArtists @ffrenzy71-arch @ceresnam @JesperBeltman @avsmithy @ksrt12 @ArtefactoO @dromadaire54 what do you think about this PR ?

patsevanton:

Mokto:
Why is it breaking ?

patsevanton:
this changes the logic and can potentially lead to errors. so I wanted to highlight this update.

Mokto:
Then I think you need to add the info the the logs

Mokto (approved):

PR #2103, opened 2026-03-27
feat: per taskbroker & taskworker resources

mzglinski:

PR #2228, opened 2026-06-01
feat: per taskbroker & taskworker topologySpreadConstraints

Llyth:
## Problem

Taskbroker and taskworker pods are using a common topologySpreadConstraints, therefore it is impossible to specify per broker or worker their own component name as a `topologySpreadConstraints.labelSelector.matchLabels`.

## Fix

Added an optional `(taskBroker|taskWorker).(brokers|workers).*.topologySpreadConstraints` to taskbroker and taskworker (guarded by check).

Based on the work made for resources in https://github.com/sentry-kubernetes/charts/pull/2103 .

### DIFF
--- a/charts/sentry/README.md
+++ b/charts/sentry/README.md
@@ -776,7 +776,7 @@ Note: this table is incomplete, so have a look at the values.yaml in case you mi
 | sentry.subscriptionConsumerTransactions.topologySpreadConstraints | list | `[]` |  |
 | sentry.subscriptionConsumerTransactions.volumes | list | `[]` |  |
 | sentry.taskBroker.affinity | object | `{}` | |
-| sentry.taskBroker.brokers | list | (see `values.yaml`) | One broker StatefulSet per item (`name`, `topic`, `consumerGroup`, `replicas`, optional `resources` merged with `sentry.taskBroker.resources`, optional `topologySpreadConstraints` merged with `sentry.taskBroker.topologySpreadConstraints`). |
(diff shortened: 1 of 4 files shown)


Llyth:
Hi @Mokto,

Copilot is right, it's based on resources' previous work and is incorrect as it's talking about merging but it's overriding/superseding.

Should topologySpreadConstraints in specific broker/worker override/supersede the default "global" taskBroker/taskWorker setting or concat ?

- Overriding would give priority to the specific setting per broker/worker.
- Concat would give priority to the strictest setting in similar key/selector.

I'll update accordingly to suit your opinion.

In my opinion, there is no perfect replication of the resources' work for topologySpreadConstraints as resources is a dict and topologySpreadConstraints is a list.

Best regards.

Mokto:
Think it should override!

Llyth:
> Think it should override!

I've made the changes in https://github.com/sentry-kubernetes/charts/pull/2228/commits/0d828b3c54ff6bb8fc706b45aab6f9af8e0f131b .

**I made sure even an empty list in the per broker/worker overrides the global setting**

| `sentry.taskBroker/Worker.topologySpreadConstraints` | `per broker/worker.topologySpreadConstraints` | Result |
|---|---|---|
| Set | Not declared | Global value is used |
| Not set (use chart default "[]") | Not declared | No topology spread constraints (empty list) |
| Set | `[]` (empty list) | `[]`, global is **ignored** |
| Set | `[...]` (non-empty) | Per-instance value is used, global is **ignored** |
| Not set | `[...]` (non-empty) | Per-instance value is used |

Mokto (approved):

PR #2233, opened 2026-06-06
feat: add startup probe for kafka consumer workloads; tighten liveness timing

vitka:
Followup to #2230. Add startup probe. Tighten liveness defaults now that startup owns cold-start.

Resolves the long-standing request in #1874 and makes the `initialDelaySeconds: 30` workaround from #2131 unnecessary.

### Startup probe

`test -f /tmp/health.txt`. Passes once arroyo has produced the file at least once. While it's running, kubelet disables liveness, so cold-start latency can't trip a restart. Defaults give ~5min cold-start window.

### Liveness timing

- `initialDelaySeconds` removed since startup probe owns cold-start; kubelet ignores liveness until startup passes
- `periodSeconds` changed from `320` to `30` for faster detection
- `failureThreshold` set to `5` - with new `periodSeconds` gives 150s slow-batch tolerance
- `timeoutSeconds` changed from `3` to `5` - headroom in case `sh + date + stat` invocation exceeds 3s under CPU throttling

Net kill latency for a deadlocked consumer should drop from ~960s to ~210s.

### --healthcheck-file flag

Previously gated on `livenessProbe.enabled`. Now always enabled. Avoids situation where disabling liveness would silently break the new startup probe.

### DIFF
--- a/charts/sentry/ci/kind-values.yaml
+++ b/charts/sentry/ci/kind-values.yaml
@@ -14,6 +14,15 @@ kafka:
     replicaCount: 1
   broker:
     replicaCount: 1
+  # The CI cluster runs only two Kafka nodes (1 controller + 1 broker), but the
+  # internal __consumer_offsets and transaction-state topics default to
+  # replication factor 3. With fewer than 3 brokers these topics cannot be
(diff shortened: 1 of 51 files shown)


Mokto (approved):
nice

PR #2236, opened 2026-06-09
fix: crash on sentry 26.5.2 due to snuba 26.5.2

lfritz-seweurodrive:
## Problem

Snuba 26.5.2 requires a new configuration to be set in the `settings.py` for snuba.
When not setting it, Snuba crashes on startup causing the Snuba Pods to remain within a CrashLoop.

## Fix

Add the configuration `events_analytics_platform_ro` to `settings.py` for snuba.


lfritz-seweurodrive:
@Mokto Same PR as previously approved one, just with conventional commit being followed

PR #2257, opened 2026-07-17
feat!: upgrade sentry to v26.7.0

mzglinski:
## Summary

Breaking chart release targeting [Sentry 26.7.0](https://github.com/getsentry/self-hosted/releases/tag/26.7.0), porting the self-hosted TaskBroker Kafka config migration and related consumer changes.

- **TaskBroker**: replace legacy `TASKBROKER_KAFKA_TOPIC` / `CONSUMER_GROUP` env vars with per-broker ConfigMaps (`kafkaTopics`, `kafkaDeadletterTopic`, `kafkaRetryTopic`) and `TASKBROKER_KAFKA_CLUSTERS__DEFAULT__*` cluster env vars ([migration guide](https://github.com/getsentry/taskbroker/blob/main/docs/kafka-config-migration.md))
- **Removed Sentry consumers** (work moves to TaskBroker `raw` topics): `subscriptionConsumer*` and `ingestProfiles` — remove any values overrides for these keys. Snuba `subscriptionConsumer*` schedulers are unchanged
- Default topic placement: subscription results → **products** broker; profiles → **ingest** broker
- Chart-wide `kafkaMaxPollIntervalMs: 300000` (self-hosted parity); per-consumer override still supported
- Feature flags: workflow-engine UI + transactions→spans migration banners; prune dead/graduated `SENTRY_FEATURES` entries that no longer exist in Sentry 26.7.0
- Upgrade notes: see `CHANGELOG.md` → **Upgrading to Chart 33.0.0** (ingress-nginx `proxy-read-timeout: "90"` when applicable)

---

Summary generated by Cursor and verified by the author of the PR.

Issue #214, opened 2020-10-20
Support .spec.strategy block in sentry-web to update correctly with persistent volumes.


sklusz:
When sentry-web is updated with `helm upgrade --install`, the sentry-web deployment spins up another pod, as it's set to a rolling deployment strategy by default.

Two web pods are created - 
```
default       sentry-web-1234                             1/1     Running             0          3d23h
default       sentry-web-1235                             0/1     ContainerCreating   0          4m36s
``` 

The sentry-web deployment strategy:
```
Replicas:               1 desired | 1 updated | 2 total | 1 available | 1 unavailable
StrategyType:           RollingUpdate
MinReadySeconds:        0
RollingUpdateStrategy:  25% max unavailable, 25% max surge
```

This leads to a timeout if using a persistent volume in the deployment, as the volume can't be mounted to the new pod while it is mounted to the old one. This is the case in AWS EKS, and I assume the same on other platforms. 

Logs for the new pod - 
```
  Normal   TriggeredScaleUp    4m41s  cluster-autoscaler       pod triggered scale-up: [{eks-... 10->11 (max: 16)}]
  Warning  FailedAttachVolume  30s    attachdetach-controller  Multi-Attach error for volume "pvc-..." Volume is already used by pod(s) sentry-web-1234
```

It is possible to make the persistent volume multi-attach using the cli, but that may cause issues with sentry (two pods writing/reading from the same volume).

This can be more easily resolved by allowing an optional `.spec.strategy` block to change the `.spec.strategy.type` to Recreate, which would bring sentry-web down prior to spinning up a new pod for the deployment. Another option is automatically changing the `.spec.strategy.type` to recreate if `.Values.filestore.filesystem.persistence.enabled`.

The only way to deploy an update while using persistence without these changes is to delete the sentry-web deployment prior to running `helm upgrade --install`.

Mokto:
Fixed in 6.2.1.
You can now change strategy type.

PR #2258, opened 2026-07-18
fix: use Recreate strategy for single-partition Kafka consumers

### LINKED ISSUE #2238: New Health&Ready check, causes commit-log deployments to never be ready
### Issue submitter TODO list

- [x] I've searched for an already existing issues [here](https://github.com/sentry-kubernetes/charts/issues)

### Describe the bug (actual behavior)

Certain consumers such as sentry-snuba-subscription-consumer-metrics, consume consume 'commit-log' events.
Those events, in sentry self hosted can only be consumed by 1 consumer at the time. If 2 attempt to consume it, the 2nd will never be ready, due there only being a single partition for the commit-log topic.

This causes, on helm update, deployments such as sentry-snuba-subscription-consumer-metrics to never be ready.

### Expected behavior

There should only be one consumer for commit-log on helm update. Otherwise, the logic of this helm chart must be changed, to maintain commit-logs in a way that allows them to be consumed by more partitions

### values.yaml

none needed

### Helm chart version

latest

### Steps to reproduce

Host the helm chart
Deploy a change affecting consumers which are commit-log
Watch & see the deployment fail, as the new pods will never be ready

alternatively:
try hosting 2 consumers of type commit-log, such as sentry-snuba-subscription-consumer-metrics

### Screenshots

_No response_

### Logs

_No response_

### Additional context

_No response_

abdalazizmoh:
## Problem

Since #2233 added healthcheck-file startup/liveness probes, any `RollingUpdate` permanently wedges the single-partition consumer deployments (monitors-clock-tick/tasks, snuba commit-log subscription consumers):

1. These topics have 1 partition consumed by 1 consumer group; the old pod holds the only assignment.
2. The surge pod gets no partition assignment, never touches `/tmp/health.txt`, fails its startup probe and is killed in a loop.
3. The Deployment never scales down the old ReplicaSet because the new pod never becomes Ready → permanent deadlock. Only manual `kubectl delete rs <old>` unblocks it.

These consumers are single-partition/single-consumer **by design** (getsentry/snuba#5855), so a surge pod can never succeed.

Fixes #2238

## Solution

Default the affected Deployments' `strategy.type` to `Recreate` **only when `replicas=1`** (where `RollingUpdate` deadlocks); otherwise default stays `RollingUpdate`. A new `sentry.kafkaConsumer.strategyType` helper in `_helper.tpl` encodes this, and an explicit `<component>.strategyType` (matching the existing `sentry.web.strategyType` idiom) always overrides either default — including for `monitorsClockTick`/`monitorsClockTasks`, which also factor in `autoscaling.enabled` since HPA can move them past 1 replica.

Implements the approach suggested by @fcuello-fudo in the #2238 discussion. Precedent: the same deadlock mechanism (surge pod unable to acquire an exclusive resource) was fixed the same way for sentry-web on ReadWriteOnce PVCs in #216 (#77, #214).

- 10 deployment templates: `monitors-clock-{tick,tasks}`; snuba `subscription-consumer-{events,transactions,metrics,generic-metrics-{counters,distributions,sets,gauges},eap-items}`
- `_helper.tpl`: added `sentry.kafkaConsumer.strategyType`
- `values.yaml`: `strategyType` now documented as `null` (dynamic default) for each block
- README: updated rows for keys already present in the parameter table

Note: this PR originally also touched 5 `sentry.subscriptionConsumer*`/`subscriptionConsumerResultsEapItems` deployments, which were removed in the meantime by the TaskBroker migration in #2257. Rebased onto `develop` and dropped those; the fix now only covers consumers that still exist.

## Scope

With `asHook: true` (default) these Deployments are recreated by the Helm hook lifecycle anyway (see #2256), so this change is a no-op there. The deadlock bites installs using `asHook: false` or GitOps tooling (e.g. Argo CD) where a real RollingUpdate happens — this fixes those.

## Verification

- `helm lint` passes
- `helm template` (default values): all 10 affected Deployments render `strategy.type: Recreate`
- Bumping a consumer's `replicas` above 1 renders `strategy.type: RollingUpdate` for that consumer, others unaffected
- `--set snuba.subscriptionConsumerEvents.strategyType=RollingUpdate` overrides back regardless of `replicas`
- `--set sentry.monitorsClockTasks.autoscaling.enabled=true` renders `RollingUpdate` despite the default `replicas: 1`

### DIFF
--- a/charts/sentry/README.md
+++ b/charts/sentry/README.md
@@ -1056,7 +1056,7 @@ Note: this table is incomplete, so have a look at the values.yaml in case you mi
 | snuba.subscriptionConsumerEvents.replicas | int | `1` |  |
 | snuba.subscriptionConsumerEvents.resources | object | `{}` |  |
 | snuba.subscriptionConsumerEvents.securityContext | object | `{}` |  |
-| snuba.subscriptionConsumerEvents.strategyType | string | `"Recreate"` | single-partition consumer: RollingUpdate surge pods deadlock the rollout |
+| snuba.subscriptionConsumerEvents.strategyType | string | `nil` | Recreate if replicas=1, else RollingUpdate |
 | snuba.subscriptionConsumerEvents.topologySpreadConstraints | list | `[]` |  |
 | snuba.subscriptionConsumerMetrics.affinity | object | `{}` |  |
 | snuba.subscriptionConsumerMetrics.containerSecurityContext | object | `{}` |  |
@@ -1069,7 +1069,7 @@ Note: this table is incomplete, so have a look at the values.yaml in case you mi
 | snuba.subscriptionConsumerMetrics.replicas | int | `1` |  |
 | snuba.subscriptionConsumerMetrics.resources | object | `{}` |  |
 | snuba.subscriptionConsumerMetrics.securityContext | object | `{}` |  |
-| snuba.subscriptionConsumerMetrics.strategyType | string | `"Recreate"` | single-partition consumer: RollingUpdate surge pods deadlock the rollout |
+| snuba.subscriptionConsumerMetrics.strategyType | string | `nil` | Recreate if replicas=1, else RollingUpdate |
 | snuba.subscriptionConsumerMetrics.topologySpreadConstraints | list | `[]` |  |
 | snuba.subscriptionConsumerTransactions.affinity | object | `{}` |  |
 | snuba.subscriptionConsumerTransactions.containerSecurityContext | object | `{}` |  |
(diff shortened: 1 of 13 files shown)


patsevanton:

Mokto (commented):
I don't think we should change the strategyType to recreate for every body.

That's a breaking change

Mokto:
:wave: Hi, @abdalazizmoh,

abdalazizmoh:
Thanks for the review, @Mokto!

- **Conflicts**: rebased onto the latest `develop` (post the v26.7.0/TaskBroker migration in #2257) and force-pushed. That migration removed 5 of the `sentry.subscriptionConsumer*` deployments this PR originally touched (`subscriptionConsumerEvents`, `subscriptionConsumerTransactions`, `subscriptionConsumerMetrics`, `subscriptionConsumerGenericMetrics`, `subscriptionConsumerResultsEapItems`) — they no longer exist, so I dropped the changes for them. The fix now only covers the deployments that are still present: `monitorsClockTasks`, `monitorsClockTick`, and the 8 `snuba.subscriptionConsumer*` consumers.
- **"Recreate for every body" / breaking change**: pushed a follow-up commit that scopes the default to `replicas=1`. There's now a `sentry.kafkaConsumer.strategyType` helper (`_helper.tpl`) that only defaults to `Recreate` when a consumer is running exactly 1 replica (the case that actually deadlocks under `RollingUpdate`); with more than 1 replica — or autoscaling enabled for `monitorsClockTasks`/`monitorsClockTick` — it defaults to `RollingUpdate` like before. An explicit `strategyType` still overrides either default.

Verified with `helm lint`/`helm template`: default install renders `Recreate` for all 10 affected consumers, bumping any one's `replicas` flips it back to `RollingUpdate`, and an explicit `strategyType` override wins regardless of `replicas`.

Mokto (approved):
Sounds good thanks!

mzglinski:
Hi, @abdalazizmoh, may I ask why you only covered some of the Sentry consumers here? Others have the same issue too.

abdalazizmoh:
Good catch, you're right, and I've opened a follow-up: #2269

The ~32 remaining consumers (ingest-*, post-process-forwarder-*, the per-entity snuba consumers, etc.) all default to `replicas: 1` and use the same execHealthcheckFile probe, so they hit the identical deadlock, the distinguishing factor was never "is this topic contractually single-partition," just "is this Deployment currently at replicas: 1", which is what the strategyType helper from this PR already handles. That follow-up wires the same helper into all of them.

PR #2269, opened 2026-07-30
fix: extend Recreate/RollingUpdate strategy to remaining Kafka consumer deployments

abdalazizmoh:
## Context

Follow-up to #2258, which defaulted `strategy: Recreate` on 10 single-partition consumer Deployments (monitors-clock-tick/tasks, snuba subscription-consumer-*) to fix the rollout deadlock described in #2238: a RollingUpdate surge pod can never get a Kafka partition assignment (the old pod already holds it), so it never touches the healthcheck file, fails its startup probe, and is killed in a loop while the old ReplicaSet is kept alive.

As raised in https://github.com/sentry-kubernetes/charts/pull/2258#issuecomment-5089626258, that scope was too narrow. The other ~32 consumer Deployments using the same `execHealthcheckFile` startup/liveness probe also default to `replicas: 1`, and the chart's own Kafka topic provisioning defaults to `numPartitions: 1` unless overridden, so on a default/minimal install (e.g. `ci/kind-values.yaml`, or any install that hasn't explicitly sized up partitions) they hit the exact same deadlock. The distinguishing factor isn't "is this topic contractually single-partition," it's simply "is this Deployment currently running `replicas: 1`" - which is what the `sentry.kafkaConsumer.strategyType` helper from #2258 already encodes.

## Changes

Wires the existing `sentry.kafkaConsumer.strategyType` helper (`Recreate` when `replicas == 1` and autoscaling is off, else `RollingUpdate`; explicit `strategyType` always wins) into the remaining 32 file-probe consumer Deployments:

- sentry: ingestConsumerAttachments, ingestConsumerEvents, ingestConsumerTransactions, ingestFeedback, ingestMonitors, ingestOccurrences, ingestReplayRecordings, billingMetricsConsumer, metricsConsumer, genericMetricsConsumer, postProcessForwardErrors, postProcessForwardIssuePlatform, postProcessForwardTransactions, processSegments, processSpans, uptimeResults
- snuba: consumer, eapItemsConsumer, genericMetricsCountersConsumer, genericMetricsDistributionConsumer, genericMetricsGaugesConsumer, genericMetricsSetsConsumer, groupAttributesConsumer, issueOccurrenceConsumer, metricsConsumer, outcomesBillingConsumer, outcomesConsumer, profilingChunksConsumer, profilingFunctionsConsumer, profilingProfilesConsumer, replaysConsumer, transactionsConsumer

Deliberately **not** touched: `snuba-replacer` (no healthcheck-file probe at all) and `snuba-outcomes-accepted-consumer` (template explicitly notes it doesn't write the healthcheck file yet) - neither can hit this deadlock. `sentry-web` and `relay` already manage their own `strategyType` independently and are unaffected.

## Verification

- `helm lint` passes
- `helm template` (`feature-complete` profile): all 41 file-probe consumer Deployments now render a `strategy` block - 39 at `Recreate` (default `replicas: 1`), `sentry-web`/`relay` unaffected at `RollingUpdate`
- `--set <component>.strategyType=RollingUpdate` overrides correctly
- Bumping any covered component's `replicas` above 1 (without an explicit override) correctly flips its default back to `RollingUpdate`

### DIFF
--- a/charts/sentry/templates/sentry/ingest/attachments/deployment-sentry-ingest-consumer-attachments.yaml
+++ b/charts/sentry/templates/sentry/ingest/attachments/deployment-sentry-ingest-consumer-attachments.yaml
@@ -19,6 +19,8 @@ metadata:
   {{- end }}
 spec:
   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  strategy:
+    type: {{ include "sentry.kafkaConsumer.strategyType" (dict "strategyType" .Values.sentry.ingestConsumerAttachments.strategyType "replicas" .Values.sentry.ingestConsumerAttachments.replicas "autoscaling" .Values.sentry.ingestConsumerAttachments.autoscaling) }}
   selector:
     matchLabels:
       app: {{ template "sentry.fullname" . }}
--- a/charts/sentry/templates/sentry/ingest/events/deployment-sentry-ingest-consumer-events.yaml
+++ b/charts/sentry/templates/sentry/ingest/events/deployment-sentry-ingest-consumer-events.yaml
@@ -19,6 +19,8 @@ metadata:
   {{- end }}
 spec:
   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  strategy:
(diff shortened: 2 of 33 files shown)


Mokto (approved):

Mokto:
Thank you!

PR #2270, opened 2026-07-31
fix: clickhouse cleanup cronjob fails when CLICKHOUSE_PASSWORD is unset

timursaikaliev:
When `externalClickhouse.password` is left at its default and `externalClickhouse.existingSecret` is not used, the `-clickhouse-cleanup` CronJob fails on startup.

Two issues:

1. The `secretKeyRef` pointed at `{{ fullname }}-clickhouse` / `clickhouse-password` — the ClickHouse subchart secret, which doesn't exist in an external-ClickHouse setup. The actual password lives in `{{ fullname }}-snuba-env` / `CLICKHOUSE_PASSWORD`.
2. Because the key is marked `optional: true`, the env var is not created at all (rather than being empty), so under `set -euo pipefail` the first reference `[ -n "$CLICKHOUSE_PASSWORD" ]` aborts the job with `unbound variable`.

Fix: point the default branch at the `-snuba-env` secret, and default the variable with `CLICKHOUSE_PASSWORD="${CLICKHOUSE_PASSWORD:-}"` before use. The `existingSecret` branch is unchanged. Verified with `helm template` in both branches.

### DIFF
--- a/charts/sentry/templates/snuba/cleanup/cronjob-clickhouse-cleanup.yaml
+++ b/charts/sentry/templates/snuba/cleanup/cronjob-clickhouse-cleanup.yaml
@@ -79,6 +79,8 @@ spec:
                 echo "Connecting to ClickHouse at $CLICKHOUSE_HOST:$CLICKHOUSE_PORT"
 
                 # Build clickhouse-client arguments; omit --password if empty to avoid bad-arg error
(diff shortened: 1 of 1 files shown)


timursaikaliev:
@Mokto, could you please take a look?

Mokto (approved):

PR #2062, opened 2026-02-28
fix: make S3 access_key/secret_key conditional for IRSA/Pod Identity support

elafkaihi:
## Problem

The filestore, replay, profiles, and nodestore S3 backends unconditionally emit `access_key`/`secret_key` 
in the generated `sentry.conf.py` via `os.getenv()` with an empty string default.

When using **IRSA** (EKS) or **Pod Identity** (EKS/AKS) — where no static credentials are 
provided — this empty string breaks the boto3 credential chain. Boto3 sees `""` as "credentials 
were explicitly provided" and stops looking for the IAM role token, causing authentication failures.

## Fix

Wrap `access_key`/`secret_key` lines in all 4 S3 backends with a conditional that checks for 
**either** `accessKey` (static credentials) **or** `existingSecret` (Kubernetes Secret):

```diff
+ {{- if or .Values.filestore.s3.accessKey .Values.filestore.s3.existingSecret }}
+ {{- if or $replayS3.accessKey $replayS3.existingSecret }}
+ {{- if or $profilesS3.accessKey $profilesS3.existingSecret }}
+ {{- if or $nodestoreS3.accessKeyId $nodestoreS3.existingSecret }}

PR #2279, opened 2026-08-17
feat: charts/sentry: source S3 and integration credentials from a Secret

kingsukhoi:
## Problem

When a credential is supplied as a plaintext value in `values.yaml`, the chart
renders it into the `<release>-sentry` ConfigMap rather than a Secret:

- `filestore.s3`, `replay.storage.s3`, `filestore.profiles.s3` and `nodestore.s3`
  access keys are baked in as the default argument of `os.getenv(...)` in
  `sentry.conf.py`.
- `github.privateKey` / `webhookSecret` / `clientId` / `clientSecret`,
  `google.*`, `slack.*` and `discord.*` are written directly into `config.yml`.

A ConfigMap is readable by anything with namespace read access, and shows up in
`helm get manifest`, GitOps diffs and cluster backups. The `existingSecret` path
already avoids this for every one of these credentials — this PR makes the
plaintext path converge on it.

## Change

- **`templates/sentry/secret-sentry-env.yaml` (new)** — a `<fullname>-sentry-env`
  Secret holding only the credentials supplied as plaintext. Rendered only when
  there is at least one, so installs entirely on `existingSecret` get no new object.
- **`templates/_helper.tpl`** — adds `sentry.credentials.sentryEnv.data`
  (the Secret contents), `sentry.credentials.sentryEnv.enabled` (the shared guard),
  `sentry.envFrom` and `sentry.config.checksum`.
- **`templates/sentry/_helper-sentry.tpl`** — drops the integration credentials from
  `config.yml`, widens the `SENTRY_OPTIONS` guards so plaintext users reach the same
  `os.environ.get(...)` lines `existingSecret` users already use, and removes the
  plaintext defaults from the eight S3 `os.getenv` calls.
- **23 workload and hook templates** — gain `envFrom` for the new Secret and switch
  their `checksum/config` annotation to `sentry.config.checksum`. 22 go through the
  `sentry.envFrom` helper; `sentry-web` already had an `envFrom` block, so it merges an
  inline `secretRef` into it instead.

The `{{- if or ... }}` guards around the S3 options are deliberately preserved, so
that when nothing is configured the keys stay absent and boto3 continues to fall
back to IRSA / Workload Identity (#2062).

`sentry.config.checksum` hashes the ConfigMap content together with the Secret data.
Without it, moving a credential out of the ConfigMap would mean rotating it no longer
changes the annotation, so pods would never restart and the rotation would silently
not take effect. Verified by rotating a credential that now lives in the Secret: the
`checksum/config.yaml` annotation changes while the ConfigMap's own hash does not.

The `secretRef` is marked `optional: true`. The Secret exists if and only if at least
one credential is chart-managed, and every consumer already tolerates these variables
being absent — the `SENTRY_OPTIONS` guards omit the option entirely and the S3 reads
default to `""`. This also matters for `hooks.preUpgrade: true`, where the two hook
Jobs run as `pre-upgrade` hooks, before Helm applies the release's normal resources:
without it, the first upgrade onto this version would fail with the hook pods stuck in
`CreateContainerConfigError`.

## Backwards compatibility

No `values.yaml` change: no new keys, none removed, no defaults altered.

The Secret and the `envFrom` are guarded on the same predicate, so for an install
that supplies every credential through `existingSecret`, neither is emitted and the
checksum reduces to `sha256(config)` — the same digest, not merely an equivalent one.
Rendered output is byte-identical and there is no pod rollout on upgrade.

Verified by rendering the chart before and after across three configurations —
every credential via `existingSecret`, no credentials configured at all, and the
full credential surface via `existingSecret` — and diffing per Kubernetes object:

```
PASS   existing-secret: byte-identical to baseline
PASS   no-credentials: byte-identical to baseline
PASS   all-existing: byte-identical to baseline
```

Installs using plaintext values roll once, as the credential moves out of the
ConfigMap and the pod spec gains an `envFrom`.

## Scope

This PR covers the S3 and integration credentials only. Counting credential values
that appear outside a Secret object, a representative install goes from 68 to 59, and
an install exercising the full credential surface from 194 to 176.

The Sentry ConfigMap is the interesting one: on a representative install it goes from
9 such values to **zero**, and on the full credential surface from 21 to 3. Those
last 3 are Redis and Kafka, which are out of scope here.

The remainder are Postgres, mail, Redis and Kafka SASL. They are deliberately left
for separate PRs rather than making this one larger.

## This is the first of a series — please merge it before the rest

Three more PRs follow, in this order:

2. **Postgres, mail and admin credentials.** ~6 files. Takes a representative install
   to 0 remaining values.
3. **Redis and Kafka SASL.** ~8 files, and the fiddliest — Redis has three parallel
   value paths (`redis.password`, `redis.auth.password`, `externalRedis.password`) with
   three different key conventions. Largest single win of the series: Redis and Kafka
   are the majority of what is left. Takes the full surface to 0.
4. **`existingSecret` support for GeoIP.** `geodata.*` is plaintext-only today. The
   value already goes to `-geoip-env`, so this adds the opt-out, not a leak fix. It is
   independent of this PR and ready whenever you want it.

**This PR needs to land before 2 and 3.** They add keys to the `-sentry-env` Secret and
build on the four helpers introduced here, so they do not stand alone. I will cut each
one from `develop` after the previous merges rather than stacking branches, to match
how this repo sequences work — so nothing else will be opened against this branch.

A fifth was planned — creating the `-clickhouse` Secret that the ClickHouse cleanup
CronJob references but the chart never creates. #2270 landed the same fix
independently while this was in review, so it is dropped and I have closed my #2280.

## Notes for review

- **Also validated on a live cluster**, not only by render-diff. Deployed to a
  production self-hosted install that uses GCS filestore with an S3 nodestore, and
  configures GitHub, Slack and Google as plaintext values. The generated Secret held
  exactly the ten expected keys; those credentials were gone from the ConfigMap while
  `github-app.id` and `github-app.name` remained; `sentry.conf.py` read exactly those
  ten names; and all ten were present in the running pods. The fleet rolled once on
  upgrade and has since run three days with zero restarts across 98 pods.
  The S3 credential path was exercised end to end rather than merely rendered: a load
  test through that install wrote new event bodies to the S3 nodestore, verified by
  fetching the resulting object out of the bucket at its expected
  `md5("<project_id>:<event_id>")` key.
  Not exercised there: Discord, and the S3 filestore / replay / profiles paths, which
  that install does not use — those remain render-verified only.
- **Partially configured integrations are still ignored**, exactly as before: Slack
  requires all three values, Discord all four, Google both. That gating is unchanged
  and the new Secret mirrors it field for field, but it is now slightly harder to
  notice, because there is no longer a `config.yml` key to eyeball.
- **`templates/sentry/metrics/deployment-metrics.yaml` is intentionally not swept.**
  It carries a `checksum/config.yaml` annotation but is the statsd exporter — it
  includes no `sentry.env` and mounts no config — so it is the one template of 24
  left on the old checksum.
- **The mechanical sweep is the bulk of the diff**: 22 of the 26 files are the same
  two-line edit. Happy to split it into its own PR ahead of this one if you would
  rather review it separately — on its own it changes no rendered output at all.
- **Synced with `develop` at 86be27f** to clear the conflict flag. The only conflicts
  were `ingest-replay-recordings` and `billing-metrics-consumer`, which v26.8.0
  deleted and which this branch had touched only with the two-line sweep — so the
  deletions were taken as-is and the sweep now covers 23 templates rather than 25.
  All figures above were re-measured against the new base after the merge.

---
🤖 Authored with [Cursor](https://cursor.com). Reviewed by me, and validated on the
live cluster described above before opening.


Mokto:
:wave: Hi, @kingsukhoi,

kingsukhoi:
Synced with `develop` at 86be27f — GitHub now reports this as mergeable, so the `conflicts` label is stale and can be dropped (I don't have permission to remove it myself).

The only conflicts were `ingest-replay-recordings` and `billing-metrics-consumer`, which v26.8.0 deleted and which this branch had touched only with the mechanical two-line sweep, so I took the deletions as-is. The sweep now covers 23 templates rather than 25.

I re-recorded the render baseline against the new `develop` and re-measured everything, since the old numbers were taken against a chart that no longer exists — the body is updated throughout. The three backwards-compatibility scenarios are still byte-identical, and the leak reductions are unchanged at 9 and 18, which is what confirms nothing was lost in the resolution.

Also closed #2280 from the series roadmap: #2270 landed the same ClickHouse fix independently while this was in review.

---

🤖 Conflict resolution and re-verification done with [Cursor](https://cursor.com), reviewed by me.

kingsukhoi:
Heads up on sequencing: the rest of this series is built and waiting on this PR.

The follow-ups are written and verified locally, but not opened yet:

- **PR 2** — Postgres / mail / admin credentials (6 files)
- **PR 3** — Kafka SASL (9 files)
- **PR 4** — Redis password (7 files)

(GeoIP `existingSecret` was independent of this one and already landed as #2284.)

One correction to the roadmap in the description above: I originally announced Redis and Kafka SASL as a single PR. They split cleanly — each independently touches the Sentry, snuba, relay and kafka-provisioning surfaces, so combined it was ~16 files. Current numbering is PR 2 = Postgres/mail/admin, PR 3 = Kafka SASL, PR 4 = Redis, PR 5 = GeoIP (merged as #2284).

How would you prefer I hand these over?

1. **Wait** — I hold them until this merges, then cut each one fresh from `develop` and open it sequentially. This is what the description currently promises.
2. **Stack them** — each opened on top of the previous, so you can see the whole credential surface at once. Downside is the bases need retargeting as they merge.
3. **Open them all now against `develop`** — you handle the merge order. Worth knowing they aren't textually independent of this PR: PR 2 also touches `_helper.tpl`, so whichever lands second will need a resync.

Whichever makes review easiest for you — just say which and I'll do that.

---

Mokto (approved):
Thank you

PR #2244, opened 2026-06-24
feat: add existingSecret support for launchpad RPC shared secret

shcherbak:
Adds launchpadTaskWorker.existingSecret/existingSecretKey, mirroring the existing sentry.existingSecret pattern, so the launchpad RPC shared secret can be sourced from an externally-managed Secret instead of the chart generating/looking it up itself.

PR #2284, opened 2026-08-18
feat: charts/sentry: add existingSecret support for GeoIP credentials

kingsukhoi:
## Problem

`geodata.accountID`, `geodata.licenseKey` and `geodata.editionIDs` are plaintext-only.

Nothing is exposed today — the chart already writes them into a `<release>-geoip-env`
Secret — so this is a parity gap rather than a leak fix. Every other credential the
chart accepts can be supplied through an externally-managed Secret; GeoIP is the one
that cannot, which means the MaxMind licence key has to live in the values file even
for users who keep every other secret out of it.

## Change

Adds `geodata.existingSecret`. When it is set, the chart does not render
`-geoip-env`, and the geoip Job's `envFrom` points at the user's Secret instead.

- **`values.yaml`** — one new key, defaulting to `""`.
- **`templates/_helper.tpl`** — `sentry.geodata.secretName`, which mirrors the
  `launchpad.secretName` helper added in #2244, and `sentry.geodata.enabled`.
- **`templates/geoip/*`** — the Secret is suppressed when `existingSecret` is set, and
  the Job's two `secretRef` entries resolve through the name helper.
- **Guard widening, 12 call sites.** Everything GeoIP-related was gated on
  `geodata.accountID` alone: the Job, the PVC, the ServiceAccount, and the
  `volumeMounts`/`volumes` pairs in `sentry-web`, `taskworker`, `taskscheduler` and
  `relay`. Without widening these, a user on `existingSecret` alone would get a PVC and
  a Job that populates it, and no pod would mount the result.

11 files, +50/−14.

## On key names

The externally-managed Secret must use the keys `GEOIPUPDATE_ACCOUNT_ID`,
`GEOIPUPDATE_LICENSE_KEY` and `GEOIPUPDATE_EDITION_IDS`. There is no
`existingSecretKey`-style override, and I want to be upfront about why, since the chart
does offer renamable keys elsewhere (`externalKafka.sasl.existingSecretKeys`).

Those three names are not a chart invention — they are the environment contract of
`ghcr.io/maxmind/geoipupdate` itself, and the `init-geoip-conf` initContainer
interpolates those exact names when it writes `GeoIP.conf`. A Secret built for
`geoipupdate` will already use them. That makes the "my Secret uses different key
names" case much rarer here than for a generic S3 or SASL credential, so `envFrom`
with fixed keys buys real simplicity at little cost.

If that turns out to be wrong, overrides can be added later without breaking anyone:
the geoip Job is a `post-install,post-upgrade` hook with no `hook-delete-policy`, so
Helm's default `before-hook-creation` already deletes and recreates it on every
upgrade. Switching it from `envFrom` to explicit `secretKeyRef` entries would cost
nothing at that point. Happy to just do it now instead if you would rather.

## Backwards compatibility

No behavior change for anyone who does not set the new key. `geodata.existingSecret`
defaults to `""`, and with it unset every widened guard reduces to exactly the
`geodata.accountID` expression it replaced — `sentry.geodata.enabled` is
`or accountID existingSecret`.

Verified by rendering the chart at this branch and at `develop` (86be27f) and diffing:

- **GeoIP configured with plaintext values** — identical, 11248 lines each.
- **GeoIP not configured** — identical, 11118 lines each, and zero `geoip` references
  in either.
- The `helm template` output is byte-identical across the three credential
  configurations the chart supports (all-`existingSecret`, nothing configured, and the
  full credential surface), so no unrelated object moved either.

The only lines that differ between any two renders are the chart's own `randAlphaNum`
values — `kraft-cluster-id`, `postgres-password`, `rpc-shared-secret` and the launchpad
key — which are regenerated on every invocation. Those four are masked before diffing;
nothing else moves.

`helm lint`: 0 failed.

## Verification of the new path

With `--set geodata.existingSecret=my-geoip-secret` and no plaintext values:

- no `sentry-geoip-env` Secret is rendered
- the Job and the `data-sentry-geoip` PVC are present
- both `secretRef` entries name `my-geoip-secret`
- the same **8** workloads mount the `geoip` volume as on the plaintext path —
  `sentry-web`, `relay`, `taskscheduler`, the four taskworkers, and the Job itself

That last count is the point of the guard widening, so it is worth stating as an
equality: 8 on the plaintext path, 8 on the `existingSecret` path, 0 with GeoIP off.

**This is render-verified only, not exercised in a cluster.** `ct install` runs against
`ci/kind-values.yaml`, which does not enable geodata, so neither the old nor the new
GeoIP path has CI coverage — and adding one would need a real MaxMind account. Flagging
it so the green check isn't read as more than it is.

## Notes for review

- **Precedence is now documented.** Setting both `existingSecret` and the plaintext
  values makes `existingSecret` win silently, matching `github.existingSecret`. That is
  reachable mid-migration, so `values.yaml` and the README both say so, along with the
  fact that a missing or misspelled key fails the hook and therefore the upgrade,
  rather than degrading quietly.
- **Helper naming.** `sentry.geodata.*` uses the `sentry.` prefix even though the
  adjacent `launchpad.*` helpers do not. The prefix is the house style in that file by
  a wide margin; happy to drop it if you prefer local consistency over file-wide.

## Related

Part of a small series moving the chart's credentials into Secrets, which starts with
#2279 — please merge that one first so the series stays in order. This PR has no code
dependency on it and can be reviewed or merged on its own; it touches no file that
#2279 touches.

Branched from `develop` at 86be27f, so it already includes the v26.8.0 changes.

---
🤖 Authored with [Cursor](https://cursor.com), reviewed by me before opening.


Mokto (approved):
Thanks

PR #2291, opened 2026-08-20
feat: charts/sentry: source Postgres, mail and admin credentials from a Secret

kingsukhoi:
Second of the credential series, following #2279 which merged as 87bb399.

## Problem

`externalPostgresql.password`, `mail.password` and `user.password` are rendered as
plaintext `value:` entries in the pod spec of every workload that includes
`sentry.env` — around thirty Deployments, plus the `sentry-cleanup` CronJob, the
`db-init` and `user-create` hook Jobs and the pgbouncer Deployment.

A pod spec is readable by anything with namespace read access, and shows up in
`kubectl get deployment -o yaml`, `helm get manifest`, GitOps diffs and cluster
backups. Each of these credentials already has an `existingSecret` path that avoids
this; as in #2279, this makes the plaintext path converge on it.

## Change

- **`templates/_helper.tpl`** — adds `sentry.credentials.postgresMail.data` as the
  single definition of both credentials, `sentry.credentials.hooksEnv.*`,
  `sentry.hooks.envFrom` and `sentry.credentials.pgbouncerEnv.*`. The three plaintext
  `value:` branches are replaced with empty branches rather than deleted (see below).
- **`templates/hooks/secret-hooks-env.yaml` (new)** — a `<fullname>-hooks-env` Secret
  for the credentials the hook Jobs need, including the admin password.
- **`templates/pgbouncer/secret-pgbouncer-env.yaml` (new)** — a
  `<fullname>-pgbouncer-env` Secret holding only the Postgres password, kept separate
  so pgbouncer is not handed the S3 keys and integration tokens in `-sentry-env` that
  it has no use for.
- **`hooks/sentry-db-init.job.yaml`, `hooks/user-create.yaml`** — switch from
  `sentry.envFrom` to `sentry.hooks.envFrom`, which emits both `secretRef`s as one
  block because a container may only have a single `envFrom`. It reduces to exactly
  `sentry.envFrom` when no hook credential is chart-managed.
- **`pgbouncer/pgbouncer-deployment.yaml`** — gains the `envFrom` and a
  `checksum/pgbouncer-env` annotation.

The Sentry fleet needs no template changes at all: `POSTGRES_PASSWORD` and
`SENTRY_EMAIL_PASSWORD` are added to the existing `-sentry-env` Secret under the key
names those pods already read, and the `envFrom` that delivers it was added in #2279.

### Why the hook Jobs get their own Secret

With `hooks.preUpgrade: true`, `sentry-db-init` (weight 6) and `user-create`
(weight 9) run as `pre-upgrade` hooks, which Helm executes *before* it applies the
release manifest. A credential held only in an ordinary Secret would therefore be
missing on the very upgrade that introduces it — the Job would fail, the Secret would
never be applied, and a retry would hit the same wall.

`-hooks-env` is itself a `pre-install,pre-upgrade` hook at weight -5, below every Job
that reads it, so it is always in place first. That is also the weight the chart
already uses for `configmap-kafka-provisioning.yaml`.

Both Secrets are populated from the same `sentry.credentials.postgresMail.data`
helper, so the duplicated keys can never disagree.

### Why the plaintext branches are kept but emptied

`sentry.env` resolves these credentials with an if/else-if chain in which plaintext
wins over `existingSecret`. Deleting the plaintext branch outright would let the
`else if .Values.*.existingSecret` arm fire for a user who set both, silently changing
which credential is used. Each branch is therefore kept with a comment in place of the
`value:`, preserving precedence exactly.

### pgbouncer rotation

pgbouncer had no checksum annotation of any kind, so moving its password into a Secret
would have meant rotating it no longer restarted the pod — the rotation would silently
not take effect. `checksum/pgbouncer-env` is added, guarded so that installs which do
not have a chart-managed pgbouncer password see no new annotation.

## Backwards compatibility

No `values.yaml` change: no new keys, none removed, no defaults altered. `README.md`
and `Chart.yaml` are untouched.

Every new Secret and every new `envFrom` is guarded on the same predicate as the data
it carries, so an install that supplies these through `existingSecret` — or does not
set them at all — renders byte-identically and sees no pod rollout on upgrade.

Verified by rendering before and after across three configurations — every credential
via `existingSecret`, no credentials configured at all, and the full credential
surface via `existingSecret` — and diffing per Kubernetes object:

```
PASS   existing-secret: byte-identical to baseline
PASS   no-credentials: byte-identical to baseline
PASS   all-existing: byte-identical to baseline
```

Installs using plaintext values roll once as the credential leaves the pod spec.

## Scope

Counting credential values that appear outside a Secret object, measured against
`develop` at bf89329 with #2279 already merged:

| | before | after |
|---|---|---|
| representative install | 59 | **0** |
| full credential surface | 176 | 116 |

The representative install reaches zero here — Postgres (30 occurrences) and mail (29)
were all that #2279 left it. The 116 remaining on the full surface are Redis (56) and
Kafka SASL (60), both out of scope for this PR.

## Remaining series

Two left, unchanged in substance from the roadmap in #2279 but renumbered, since Redis
and Kafka SASL split into separate PRs rather than the single one originally announced
— each independently touches the Sentry, snuba, relay and kafka-provisioning surfaces,
so combined it was ~16 files.

- **Kafka SASL** — 9 files. Written and verified against this branch already; takes the
  full surface from 116 to 56.
- **Redis** — 7 files. The fiddliest of the series: three parallel value paths
  (`redis.password`, `redis.auth.password`, `externalRedis.password`) with three key
  conventions. Takes it to 0.

Both are ready. As promised in #2279, I will cut each from `develop` after the previous
merges rather than stacking branches, so nothing will be opened against this branch.
Say the word if you would rather see them all at once.

## Notes for review

- **Render-verified, not yet cluster-verified.** Unlike #2279, this one has not yet
  been through a live install. The before/after object diffs and the leak counts above
  are the evidence here. The install I validated #2279 on does run
  `postgresql.enabled: false` with a plaintext `externalPostgresql.password` and a
  plaintext `mail.password`, so it exercises exactly these paths and is where I intend
  to validate this — happy to hold the merge until I have done so if you would prefer.
- **The ArgoCD annotation on the new hook Secret** follows the existing convention in
  `hooks/sentry-secret-create.yaml` (`argocd.argoproj.io/hook: Sync` under
  `useArgoCdCompatibleAnnotations`). I see #2243 proposes dropping that annotation from
  `sentry-secret`; if it lands I will follow whatever it settles on here too.
- **`postgresql.enabled` is untouched.** When the bundled subchart is in use the
  password already comes from the subchart's own Secret, and both new helpers keep the
  `not .Values.postgresql.enabled` condition that guards it.
- **`user.existingSecret` still wins over `user.password`**, as it does today; the
  admin password only enters `-hooks-env` when no `existingSecret` is set.

---

### DIFF
--- a/charts/sentry/templates/_helper.tpl
+++ b/charts/sentry/templates/_helper.tpl
@@ -872,6 +872,25 @@ DISCORD_PUBLIC_KEY: {{ .Values.discord.publicKey | b64enc | quote }}
 DISCORD_CLIENT_SECRET: {{ .Values.discord.clientSecret | b64enc | quote }}
 DISCORD_BOT_TOKEN: {{ .Values.discord.botToken | b64enc | quote }}
 {{- end }}
+{{- include "sentry.credentials.postgresMail.data" . }}
+{{- end -}}
+
+{{/*
+The Postgres and mail credentials the chart manages itself, in the key names the
+Sentry fleet already reads. Shared by the Sentry Secret and the hook Secret so the
+two can never disagree about a value.
+
+The conditions mirror the env blocks these replace: postgresql.enabled takes the
+password from the subchart's own Secret, and mail.password takes precedence over
+mail.existingSecret.
+*/}}
+{{- define "sentry.credentials.postgresMail.data" -}}
+{{- if and (not .Values.postgresql.enabled) .Values.externalPostgresql.password }}
+POSTGRES_PASSWORD: {{ .Values.externalPostgresql.password | b64enc | quote }}
+{{- end }}
+{{- with .Values.mail.password }}
+SENTRY_EMAIL_PASSWORD: {{ . | b64enc | quote }}
+{{- end }}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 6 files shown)


Mokto:
Blocking: numeric passwords no longer render. `b64enc` errors on a non-string, where the old `| quote` accepted one:

```
$ helm template rel charts/sentry -f vals.yaml --set externalPostgresql.password=12345
Error: template: sentry/templates/_helper.tpl:889:60
  executing "sentry.credentials.postgresMail.data" at <b64enc>:
  wrong type for value; expected string; got int64
```

An unquoted `password: 12345` in `values.yaml` parses as an int, so an install that renders today would break on upgrade. Fix is `toString | b64enc` in the four new spots — `postgresMail.data` (both keys), `hooksEnv.data`, `pgbouncerEnv.data`.

(#2279 has the same latent issue on its keys, but these three values are much more commonly set, so it is likely to actually be hit here.)

Rest of the PR verifies out: rendered your three configs plus a bundled-Postgres one against bf89329 — the `existingSecret` config is byte-identical modulo the `randAlphaNum` values, and a per-container check across the plaintext config confirms no container loses a credential. Separate non-blocking comments to follow.

kingsukhoi:
Good catch, and you're right that it's a regression rather than something pre-existing. Fixed in 2bd3411 — all four spots you listed now go through `toString | b64enc`: `postgresMail.data` (both keys), `hooksEnv.data` and `pgbouncerEnv.data`.

Confirmed the before/after with a values file carrying unquoted integer passwords (`--set` parses types differently, so I used a real file):

| | result |
|---|---|
| `develop` before this PR | renders |
| this PR as you reviewed it | fails on all four |
| this PR now | renders |

The values also round-trip rather than merely rendering: `POSTGRES_PASSWORD`, `SENTRY_EMAIL_PASSWORD`, `ADMIN_PASSWORD` and `POSTGRESQL_PASSWORD` decode out of the Secret as `12345` / `67890` / `24680` / `12345`, byte-identical to the `value:` `develop` puts in the pod env for the same input.

One related thing I checked and deliberately did **not** change: a password containing a decimal point parses as a YAML float and renders as `1.23456789012345e+09`. That is exactly what `develop` does today — `| quote` mangles it identically — so `toString` restores parity here rather than introducing or fixing anything. Mentioning it only in case you think the chart should reject or string-coerce such values; it felt out of scope for this PR.

Also resynced with `develop` at 6bf2b2c (picking up the taskbroker migrate job), re-recorded the render baseline against it, and re-verified everything: the three backwards-compatibility configs are byte-identical, the representative install still goes 59 → 0, the full credential surface 176 → 116, and `helm lint` reports 0 failed.

## On your #2279 parenthetical

It's worse than latent — three keys break on `develop` right now, same values-file test:

- **`discord.applicationId`** (#2279) — Discord application IDs are snowflakes, all digits
- **`slack.clientId`** (#2279)
- **`geodata.accountID`** (#2284, also mine) — MaxMind account IDs *are* numeric, so this is the one most likely to actually be hit

Happy to open a small separate PR against `develop` for those three, since they're independent of this one. Say the word and I'll put it up.

For the rest of the series: PR 3 (Kafka SASL) is immune — its values come through `include`, which always returns a string — and PR 4 (Redis) reads `.Values.redis.password` directly, so it'll carry the same `toString` guard from the start.

Still watching for the non-blocking comments you mentioned.

---

🤖 Fix and re-verification done with [Cursor](https://cursor.com), reviewed by me.

